PR #6, opened 2020-02-10
Add design tokens for typography, spacing, color

meggles:
Add design token values for spacing, color, and typography and display them as stories. There is some lorem ipsum I've left as a reminder to fill in later.

I'm planning to refactor the redundant table styles soon.


src/components/DesignTokens/Color/Color.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  td {
+    font: 1.6rem/1.5 ${font.pnr};
+    margin-right: 1.6rem;
+  }
+
+  tr {
+    border-bottom: 2px solid #EFEFEF;
+    padding-left: ${spacing.md};
```
jaslakson: Should we create a color name for this one?

meggles: Good thinking! We'll only use this color in the storybook app (not in our exportable components), but it would be good to keep it looking consistent.

src/components/DesignTokens/Spacing/Spacing.js
```diff
@@ -0,0 +1,96 @@
...
+
+  Object.keys(tokens).forEach((token) => {
+    const tokenValue = tokens[token];
+    if (tokenValue) {
+      tokensArray.push(
+        <>
+          <td>{`spacing.${token}`}</td>
+          <td>{tokenValue}</td>
+          <td>
+            <span style={{width: `${tokenValue}`, display: 'inline-block', backgroundColor: '#c2c2c2', height: 1 + 'rem'}}></span>
```
mirdathebirda: should we pass in color variable for backgroundColor here instead of #c2c2c2?

atk-dm:

PR #11, opened 2020-02-12
made TokenTable into its own component

mirdathebirda:
starting off with something simple but more iterations (probably) to come


atk-dm:

PR #13, opened 2020-02-13
moving common components used in design tokens to a shared folder

mirdathebirda:
we were repeating TokenTable and ImportCode code a lot inside the design tokens so I moved them to a shared folder


src/components/DesignTokens/utils/index.js
```diff
@@ -0,0 +1,7 @@
+import TokenTable from './TokenTable';
+import ImportCode from './ImportCode';
+
+export {
+  TokenTable,
+  ImportCode,
+};
```
mirdathebirda: I currently moved these two components into a folder called 'utils' but I'm not sure this is the best name for it? Any other ideas? shared maybe?

meggles: What if we just called it `components`? ...Or maybe `token-components` (but maybe that;s redundant since it's in a directory called `DesignTokens`?

mirdathebirda: I think `components` is good. I changed it to that for now

atk-dm:

PR #15, opened 2020-02-14
icons

mirdathebirda:
first draft of icons component

since we don't have much information on icons right now (aside from the badge icon), leaving room for any props we might add later

we didn't get any svgs from design yet, so current svgs are just placeholders


src/components/DesignTokens/Icon/svgs/ATK.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+
+const Atk = (props) => {
```
jaslakson: with some versions of brand icons we've allowed `backgroundFill` or `fill` as a prop so that we can change the color of the icons. This is helpful for grayscale icons or icons with inverted colors.

mirdathebirda: since these are just placeholders for now, i decided to leave out the customizability of the fills. i don't think we've reached a decision yet on whether or not logos would count as icons... 🤔but will keep this info in mind!

atk-dm:

PR #17, opened 2020-02-18
swapping out the svgs for something more mutable and adding fill and stroke as props to icon

mirdathebirda:


atk-dm:

PR #18, opened 2020-02-18
remove public folder

mirdathebirda:
doesn't look like we're using anything in this folder


atk-dm:

PR #19, opened 2020-02-18
renaming files to be index.js so we don't have to keep making extra index.js files for exports

mirdathebirda:


atk-dm:

PR #21, opened 2020-02-20
remove color related props from Icon

mirdathebirda:
having them as props makes it more difficult to deal with any changes based on css (ex: hover states)


atk-dm:

PR #26, opened 2020-02-21
Breakpoints, responsive cards, icon improvements

meggles:
– add styled-components-breakpoints
– resize some card elements for mobile
– update favorite icon svg code
– make aria attributes required props for icons

### DIFF
--- a/src/components/Card/index.js
+++ b/src/components/Card/index.js
@@ -1,24 +1,37 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import breakpoint from 'styled-components-breakpoint';
 import { color, font, fontSize, lineHeight, spacing } from '../../styles';
 import { Sticker } from '../Sticker';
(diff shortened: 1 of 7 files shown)


src/components/Card/index.js
```diff
@@ -48,6 +65,17 @@ const StickerGroup = styled.div`
   bottom: 0;
 `;
 
+const stickerHeightMobile = '1.2rem';
+const StyledSticker = styled(Sticker)`
+
+  ${breakpoint('mobile', 'tablet')`
+    border-radius: 0.5rem;
+    line-height: ${stickerHeightMobile};
+    height: ${stickerHeightMobile};
```
mirdathebirda: just wondering, why are we using a variable rather than just using the value here?

meggles: I used a variable here to show that there's a relationship between the two values and that it's not by chance that they have the same value. It would break if one of the values changed.

atk-dm:

PR #30, opened 2020-02-26
Add hover state to favorite ribbon

meggles:
When hovered, the ribbon shape should get a fill and a plus should appear. We'll need another hover state at some point for removing a favorited card – the svg is built to make that very easy when the time comes.

### DIFF
--- a/src/components/DesignTokens/Icon/svgs/FavoriteRibbon.js
+++ b/src/components/DesignTokens/Icon/svgs/FavoriteRibbon.js
@@ -1,5 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { color } from '../../../../styles';
+
+const StyledFavoriteRibbon = styled.svg`
+  cursor: pointer;
+
+  [class*="ribbon"] {
+    fill: transparent;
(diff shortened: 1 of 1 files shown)


src/components/DesignTokens/Icon/svgs/FavoriteRibbon.js
```diff
@@ -1,29 +1,47 @@
...
-    <g transform="translate(1 1)">
-      <path
-        d="M7.483 15.809L0 21V0h15v21z"
-        strokeMiterlimit="10"
-        strokeWidth="2"
-      />
-    </g>
-  </svg>
+      <g strokeWidth="2">
+        <path id="ribbon" d="M8.483 16.809L1 22V1h15v21z" fill="#3d3d3d" stroke="#3d3d3d" strokeMiterlimit="10"/>
```
jaslakson: Do you think using an id will cause a problem? or, would a className be better? we'll probably have lots of these on any given page.

meggles: 💯 yes yes of course... great catch, thanks. will make that change!

atk-dm:

PR #31, opened 2020-02-26
setup for some basic jest tests

mirdathebirda:
since we already use jest in other repos (espresso) and storybook recommended jest snapshot testing for structural testing, I thought it would be good to use in mise too

for now just covering structural testing and adding more as i comb through research of different testing types


atk-dm:

PR #32, opened 2020-02-26
minor updates

mirdathebirda:
playing around with some other types of knobs
changed sticker to be a named export instead of member export

### DIFF
--- a/src/components/Sticker/index.js
+++ b/src/components/Sticker/index.js
@@ -71,23 +71,20 @@ Stickers scale down to a smaller size (`height: 1.2rem;` `font-size: 0.8rem;`) f
 
 */
 
-export function Sticker({
+const Sticker = ({
   className,
   text,
   iconType,
-  fill,
   isPriority,
-}) {
-  return (
-    <StyledSticker
-      className={className}
-      isPriority={isPriority}
-    >
-      {determineIconType(iconType, fill)}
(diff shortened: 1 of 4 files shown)


src/components/Sticker/index.js
```diff
@@ -71,19 +71,18 @@ Stickers scale down to a smaller size (`height: 1.2rem;` `font-size: 0.8rem;`) f
 
 */
 
-export function Sticker({
+const Sticker = ({
   className,
   text,
   iconType,
-  fill,
```
mirdathebirda: i don't think this fill prop is being used anywhere?

src/components/Sticker/index.js
```diff
@@ -71,19 +71,18 @@ Stickers scale down to a smaller size (`height: 1.2rem;` `font-size: 0.8rem;`) f
...
   isPriority,
-}) {
+}) => {
```
jaslakson: if you wanted, you could drop the `{}` and `return`, something like 
```
const Sticker = ({...}) => (
    <StyledSticker.../>
);
```

In Espresso we have the linter set up to recommend this approach - it would be good to see if we can standardize those settings.

mirdathebirda: ooo nifty! i'll spin up a `.eslintrc` file in another branch

atk-dm:

PR #39, opened 2020-03-02
add eslintrc

mirdathebirda:
for the most part, i kept the same rules that we use in espresso


atk-dm:

PR #43, opened 2020-03-09
search box with algolia's instant search

mirdathebirda:
currently using an example search client given in algolia's getting started setup

### DIFF
--- a/src/components/DesignTokens/Icon/svgs/SearchIcon.js
+++ b/src/components/DesignTokens/Icon/svgs/SearchIcon.js
@@ -1,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SearchIcon = ({fill}) => {
+  return (
+    <svg
+      className="search-icon"
+      role="img"
+      aria-label="search"
+      viewBox="0 0 30 29.977"
+    >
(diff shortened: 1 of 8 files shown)


src/components/DesignTokens/Icon/svgs/SearchIcon.js
```diff
@@ -0,0 +1,25 @@
...
+
+const SearchIcon = ({fill}) => {
+  return (
+    <svg
+      className="search-icon"
+      role="img"
+      aria-label="search"
+      viewBox="0 0 30 29.977"
+    >
+      <defs><clipPath id="a"><path data-name="Path 83" d="M1049.75-79.754a7.5 7.5 0 017.5-7.5 7.5 7.5 0 017.5 7.5 7.5 7.5 0 01-7.5 7.5 7.5 7.5 0 01-7.5-7.5m25.595 15.38l-8.906-8.907a11.193 11.193 0 002.053-6.473A11.246 11.246 0 001057.245-91 11.246 11.246 0 001046-79.754a11.246 11.246 0 0011.245 11.246 11.194 11.194 0 006.5-2.071l8.9 8.9a2.016 2.016 0 002.833.136 2.015 2.015 0 00-.136-2.833" transform="translate(-1046 91)" fill={fill} clipRule="evenodd"/></clipPath><clipPath id="b"><path data-name="Path 82" d="M0 521.482h1600V-121H0z" transform="translate(0 121)" fill={fill}/></clipPath></defs><g data-name="Group 779" clipPath="url(#a)"><g data-name="Group 778" transform="translate(-1042.003 -25.023)" clipPath="url(#b)"><path data-name="Path 81" d="M1041.003 25.019h30.913v30.92h-30.913z" fill={fill}/></g></g>
```
meggles: I think we need to add prefixes to these ids & clip paths in svgs because when design exports them, XD uses very generic id names like `a` so the risk of a collision is high. For other icons, I prefixed it with the icon name – maybe we keep with that pattern? `clipPath id="search-icon-a">` and `clipPath="url(#search-icon-a)">`

Also I think we can remove all of the `data-name` attributes.

mirdathebirda: oo yes ty

atk-dm:

PR #44, opened 2020-03-09
Move media query from Badge to Card

meggles:
Badges don't always scale down for mobile widths


atk-dm:

PR #46, opened 2020-03-09
Create reusable algolia InstantSearch wrapper

ttornese:
wrap any component we use from algolia instant search with `MiseInstantSearch` and it will use our index and hits from that index

### DIFF
--- a/src/components/SearchInput/index.js
+++ b/src/components/SearchInput/index.js
@@ -1,11 +1,11 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import { connectSearchBox } from 'react-instantsearch-dom';
+
 import { color, font, fontSize } from '../../styles';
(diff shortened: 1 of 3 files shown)


src/components/SearchInput/index.js
```diff
@@ -83,38 +83,18 @@ const StyledSearchBox = ({ currentRefinement, refine, placeholder }) => (
...
-    >
-        <SearchBox
-          placeholder={placeholder}
-        />
-      {children}
-    </InstantSearch>
+const SearchInput = ({ placeholder }) => (
+  <SearchBox
+    placeholder={placeholder}
+  />
```
mirdathebirda: Since we're not using `<InstantSearch>` here anymore, we can get rid of this extra variable/change the naming of the other two variables (SearchBox & StyledSearchBox)

atk-dm:

PR #48, opened 2020-03-11
NPM package setup using auto

jaslakson:
I still sort-of have no idea what I'm doing with this. Following the instructions in the storybook/auto documentation - I think we need to get this setup and then merged into master. From there I can run the actual publish command. 

We'll be publishing to our @npm package space, which currently contains `pantry`. let me know if you have strong objections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,3 +10,16 @@ Mise UI is a [Storybook](https://storybook.js.org) app built with [React](https:
 
 ### Visual review
 To review work with others, push your branch to `origin/staging`. Netlify will build and deploy to http://staging--mise-ui.netlify.com. The master branch is always viewable here: https://mise-ui.netlify.com
+
+## Contributors ✨
+
+Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):
+
+<!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-disable -->
+<!-- markdownlint-enable -->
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -10,3 +10,16 @@ Mise UI is a [Storybook](https://storybook.js.org) app built with [React](https:
...
+Thanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):
+
+<!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->
+<!-- prettier-ignore-start -->
+<!-- markdownlint-disable -->
+<!-- markdownlint-enable -->
+<!-- prettier-ignore-end -->
+<!-- ALL-CONTRIBUTORS-LIST:END -->
+
+This project follows the [all-contributors](https://github.com/all-contributors/all-contributors) specification. Contributions of any kind welcome!
```
meggles: I assume this part was autogenerated.. Do we need it since we don't use prettier or markdown lint?

jaslakson: yeah - this was auto generated. I can remove it.

jaslakson: Although - it says that we shouldn't remove or modify that section, which is strange...Let's leave it for now and we can always take it out later. That's for the 'contributors' plugin, which we can decide to remove later too. I think it just says which people contributed to each release.

meggles: Ok sure, makes sense

jaslakson:
@meggles I don't think it matters. I think we can even re-publish with the same version number if we want to do that when we're approved. For now I'd just like to see if I can get it up on NPM and installed in jarvis as a package.

atk-dm:

PR #49, opened 2020-03-12
ARCHIE-286: creating SearchRefinementList component

ttornese:


src/components/SearchRefinementList/SearchRefinementList.stories.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+// Need some padding in Storybook only. This is so you can view absolutely
+// positioned elements when interacting with the filters.
+const StyledWrapper = styled.div`
+  padding: 0 0 0 3rem;
+`;
+
+export const SearchSiteList = () => {
+  const transformItems = (items) => {
```
ttornese: what does everyone think of having this transform here? what site keys and what order we want to render them seems like it could change and seems like logic that should live in jarvis. do we want to have to do a storybook package release to change the CIO button to be above ATK, etc.? that's why i hard coded an example transform here rather than having it live in the component itself.

src/components/SearchRefinementList/index.js
```diff
@@ -0,0 +1,56 @@
...
+        items.map((item) => {
+          return (
+            <SearchSiteList
+              {...item}
+              key={item.value}
+              refine={refine}
+            />
+          );
+        })
+      ) : null
```
ttornese: the next pull i do will update this `null` to be a generic refinment list. all refinement lists other than the `search_site_list` facet share the same behavior and styles

src/components/SearchRefinementList/SearchRefinementList.stories.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import SearchRefinementList from '../SearchRefinementList';
+import styled from 'styled-components';
+import { orderBy } from 'lodash';
```
jaslakson: can we do individual packages instead of the full `lodash`? As this stands, the full bundle will be shipped.

src/components/SearchRefinementList/SearchRefinementList.stories.js
jaslakson: We could start with it here - but I would like to memoize this function so that if the filters re-render a bunch of times we don't have to do the logic each time. Also, can you document what you're doing between lines 24-28 - I think I understand but I'm not sure.

src/components/DesignTokens/Icon/svgs/Checkmark.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Checkmark = ({ fill }) => (
+  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 8">
```
meggles: We might want to add some accessibility attributes here. `aria-hidden="true"` `role="img"` and `focusable="false"` – those are what we have documented in the a11y guidelines right now..

src/components/DesignTokens/Icon/svgs/Checkmark.js
```diff
@@ -0,0 +1,18 @@
...
+    <path fill={fill} d="M11.838.223a.657.657 0 00-.907-.076L3.245 6.492 1.181 4.149a.652.652 0 00-.632-.272.641.641 0 00-.522.442.621.621 0 00.186.65L2.68 7.791l.048.047c.019 0 .036.035.055.047a.325.325 0 00.055.032.689.689 0 00.065.038l.068.022a.5.5 0 00.065.019.614.614 0 00.136 0 .661.661 0 00.11 0h.016l.094-.022h.019l.084-.038.026-.019.065-.044 8.19-6.755a.621.621 0 00.065-.9z" />
+  </svg>
+);
+
+Checkmark.propTypes = {
+  fill: PropTypes.string,
+};
+
+Checkmark.defaultProps = {
+  fill: '#3D3D3D',
```
meggles: Might be good to use a variable here.

src/components/SearchRefinementList/components/SearchSiteList.js
```diff
@@ -0,0 +1,107 @@
...
+          className="search-refinement__badge"
+          fill={isRefined ? altFill : color.eclipse}
+          type={key}
+        />
+        <SearchRefinementListLabel>
+          {label}
+        </SearchRefinementListLabel>
+      </SearchRefinementListButton>
+    </SearchRefinementListItem>
+  );
```
meggles: I'm trying to think about how we can make this a11y friendly – maybe allow a screen reader to announce the selected filters ...

src/components/SearchRefinementList/SearchRefinementList.stories.js
mirdathebirda: I think it makes sense to keep this transform here (for story/visual example sense) since this logic would be placed in Jarvis. In Mise, I'm not exactly sure what protocol we should follow when stuff is based on business logic (but perhaps that's a conversation we would need to have with QA?)

src/components/SearchRefinementList/components/SearchSiteList.js
```diff
@@ -0,0 +1,109 @@
...
+    'shop': {
+      altFill: color.tomato,
+      label: 'ATK Shop',
+    },
+  }[key] || ''
+);
+
+export default ({ isRefined, label: key, refine, value }) => {
+  const { altFill, label } = keyToButton(key);
+  const onClick = (e) => {
```
jaslakson: Can we inline this on line 98 as a one-liner?

PR #58, opened 2020-03-17
ARCHIE-278: creating CurrentRefinements component and LabelFrame component

ttornese:
`LabelFrame` is a decorative component. It takes a `label` prop that will be rendered in the top left corner of the wrapper. Useful for instant search components so you can separate and organize the actual component from results.

### DIFF
--- a/src/components/LabelFrame/index.js
+++ b/src/components/LabelFrame/index.js
@@ -1,0 +1,72 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+import { color, font, fontSize } from '../../styles';
+
+const Frame = styled.div`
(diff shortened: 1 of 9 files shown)


src/components/Badge/Badge.stories.js
```diff
@@ -9,14 +11,38 @@ export default {
   decorators: [withKnobs],
 }
 
-export const ATKBadge = () => <Badge type='atk' fill={text('fill', color.transparentBlack)} />
+export const ATKBadge = () => (
+  <LabelFrame label="Component">
+    <Badge type='atk' fill={text('fill', color.transparentBlack)} />
+  </LabelFrame>
```
meggles: This is really really awesome for the search results components, but to me, it seems maybe unnecessary for other components..

PR #62, opened 2020-03-18
update list of exported packages

jaslakson:
can't import to jarvis if we don't export from here! Want to merge this to master and get up to npm, will rebase the new feature branch after.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,9 +2,13 @@ import * as styles from './styles';
 import Badge from './components/Badge';
 import Button from './components/Button';
 import StandardCard from './components/Cards/StandardCard';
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -1,10 +1,15 @@
 import * as styles from './styles';
 import Badge from './components/Badge';
 import Button from './components/Button';
+import LabelFrame from './components/LabelFrame';
```
ttornese: `LabelFrame` doesn't need to be exported. it's only used for display purposes in mise

jaslakson: got it - removing.

atk-dm:
:rocket: PR was released in `v0.1.4` :rocket:

PR #64, opened 2020-03-19
undo the connect changes, small tweaks

jaslakson:
* we *can* connect in mise-ui!
* small changes based on what is rendered in jarvis

[UPDATE]
* prop updates to `StandardCard` and `FavoriteRibbon` based on `jarvis` needs
* use `is-favorited` classname to adjust styles for `FavoriteRibbon` since this classname is set by an external script at run-time (not render time) via the favorites widget code
* debounce the search input, make it an uncontrolled component. This allows `jarvis` to not pass `searchState` down the render tree and still allows users to freely type in the input
* include `@babel/plugin-proposal-class-properties` so that we can have arrow functions on classes
* use `id` as prop instead of `key` for `ShowMoreLess` b/c `key` is a reserved prop name w/React

### DIFF
--- a/src/components/Cards/shared/FavoriteButton/index.js
+++ b/src/components/Cards/shared/FavoriteButton/index.js
@@ -26,7 +26,7 @@ const StyledFavoriteButton = styled.button`
     stroke: ${color.white};
   }
 
-  ${props => props.isFavorited && css`
+  &.is-favorited {
     [class*="ribbon"] {
       fill: ${color.eclipse};
     }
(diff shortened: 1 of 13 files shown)


src/components/Cards/shared/FavoriteButton/index.js
```diff
@@ -45,29 +45,38 @@ const StyledFavoriteButton = styled.button`
...
-const FavoriteButton = ({ isFavorited, onClick }) => (
+const FavoriteButton = ({
+  objectId,
+  onClick,
+  siteKey,
+  title,
+}) => (
   <StyledFavoriteButton
-    isFavorited={isFavorited}
+    className="favorite-action"
```
jaslakson: This className and these `data-` attributes are required for the fav widget to work properly

src/components/Cards/shared/FavoriteButton/index.js
```diff
@@ -45,29 +45,38 @@ const StyledFavoriteButton = styled.button`
...
+    data-document-title={title}
+    data-favoritable-id={objectId}
+    data-origin-site={siteKey}
     onClick={onClick}
   >
     <FavoriteRibbon
       ariaHidden="true"
-      ariaLabel={isFavorited ? 'Remove from favorites' : 'Save to favorites'}
+      ariaLabel={'Save to favorites'}
```
jaslakson: Without the `isFavorited` prop we can't change the aria label here. However, we could update the favorites widgets so this value is adjusted based on favorite status.

meggles: Should we make aria-label a prop then?

PR #68, opened 2020-03-23
Pass url to card

meggles:
also: use more realistic images, improve a11y

### DIFF
--- a/src/components/Cards/StandardCard/StandardCard.stories.js
+++ b/src/components/Cards/StandardCard/StandardCard.stories.js
@@ -12,13 +12,13 @@ export default {
 export const LoggedIn = () => (
   <StandardCard
     badgeType="atk"
-    contentType="cooking school course"
+    contentType="review"
     commentCount={5}
     ctaText={text("CTA text", "Buy the Winner")}
     ctaUrl="https://www.amazon.com/dp/B01JCNEJSO/?tag=ciosearchresult-20"
     displayFavoritesButton={true}
-    imageAlt={text("Image alt text", "Gray cat pulling up the edge of a rug")}
-    imageUrl={text("Image url", "https://placekitten.com/272/272")}
+    imageAlt={text("Image alt text", "")}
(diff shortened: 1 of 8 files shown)


src/components/Cards/StandardCard/StandardCard.stories.js
```diff
@@ -12,19 +12,20 @@ export default {
...
     badgeType="atk"
-    contentType="cooking school course"
+    contentType="review"
     commentCount={5}
     ctaText={text("CTA text", "Buy the Winner")}
     ctaUrl="https://www.amazon.com/dp/B01JCNEJSO/?tag=ciosearchresult-20"
     displayFavoritesButton={true}
-    imageAlt={text("Image alt text", "Gray cat pulling up the edge of a rug")}
-    imageUrl={text("Image url", "https://placekitten.com/272/272")}
+    imageAlt={text("Image alt text", "")}
```
meggles: Leaving this alt attribute as an empty string to hide it from screen readers as it is redundant with the card title.

src/components/Cards/StandardCard/index.js
```diff
@@ -112,45 +120,49 @@ export function StandardCard({
...
+                />
+              ))}
+            </StickerGroup>
+          : null }
+        </ImageWrapper>
+        <TitleWrapper>
+          <StyledTitle className={className} title={title} />
+          { displayFavoritesButton ? (
+            <StyledFavoriteButton
+              ariaLabel="Save to favorites"
```
mirdathebirda: would this also need to toggle between save to favorites and remove from favorites like below?

meggles: I would say yes, but I saw in Jason's PR that the `isFavorited` prop has to be removed. Sounds like there's more work on the Jarvis side that needs to happen for that. I think it's important though – otherwise the favoriting feature isn't really available for non-sighted users. I'll make a Jira card so we don't forget.

PR #74, opened 2020-03-31
Add show more results button

meggles:


src/components/ShowMoreResults/index.js
```diff
@@ -0,0 +1,35 @@
...
+  refineNext,
+  resultsCount,
+}) => {
+  return (
+    <Button
+      className={className}
+      disabled={!hasMore}
+      onClick={refineNext}
+    >
+      Show {`${resultsCount}`} more results
```
meggles: I wasn't sure if there's a way to get `resultsCount` dynamically inside of mise? Or if this logic should live in jarvis... any thoughts?

jaslakson: I think we have to pass it in as a prop from `jarvis` - looking at the documentation from `connectInfiniteHits`, I don't think we have any context regarding page size.

atk-dm:
:rocket: PR was released in `v0.1.6` :rocket:

PR #75, opened 2020-03-31
Fix reset button, which wasn't displaying

meggles:


atk-dm:
:rocket: PR was released in `v0.1.6` :rocket:

PR #80, opened 2020-04-02
Filter button & show/hide widths

meggles:
– move class from svg to button element to properly style filter button externally
– allow ShowHide and SearchSortBy components to take up the full width of their parent
– move some styles to global stylesheet
– neaten focus blur for show hide button

### DIFF
--- a/src/components/SearchSortBy/index.js
+++ b/src/components/SearchSortBy/index.js
@@ -4,15 +4,12 @@ import styled from 'styled-components';
 import { connectSortBy } from 'react-instantsearch-dom';
 
 import { ShowHide } from '../ShowHide';
-import { color, font, fontSize } from '../../styles';
+import { color, font, fontSize, spacing, } from '../../styles';
 
-const SearchSortByList = styled.ul`
-  padding: 0;
-  margin: 0;
-`;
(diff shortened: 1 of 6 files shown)


src/components/ShowHide/index.js
```diff
@@ -3,18 +3,10 @@ import PropTypes from 'prop-types';
...
-const ShowHideDivWrapper = styled.div`
-  max-width: 20rem;
-`;
-
-const ShowHideFieldsetWrapper = styled.fieldset`
-  border: 0;
-  margin: 0;
-  max-width: 20rem;
-  padding: 0;
-`;
```
meggles, line 17: I felt like these belonged in the global stylesheet, but it makes me wonder if normalize is working.. going to look into it.

src/components/SearchSortBy/index.js
```diff
@@ -21,10 +18,7 @@ const SearchSortByItem = styled.li`
 
 const SearchSortByButton = styled.button`
   align-items: center;
-  border: none;
   display: flex;
-  margin: 0;
-  padding: 0;
```
meggles, line 27: These are already applied by global.js so I removed them.

PR #94, opened 2020-04-08
ARCHIE-243: updating components to support mixpanel events being sent from jarvis

ttornese:


src/components/SearchClearRefinements/index.js
```diff
@@ -1,25 +1,59 @@
 import React from 'react';
-import { ClearRefinements } from 'react-instantsearch-dom';
+import PropTypes from 'prop-types';
+import { connectCurrentRefinements } from 'react-instantsearch-dom';
```
ttornese, line 3: this needed to be re-written with connect to support on click. and it does use `connectCurrentRefinments` not `connectClearRefinements`. this is whats outlined in the docs

PR #101, opened 2020-04-09
adjust spacing of stickers for cards with no image

jaslakson:
* provide a little bit of space and wrap stickers. This seems to work nicely for stickers that are on one line and multiple lines
* set background image wrapper to transparent when no image

### DIFF
--- a/src/components/Cards/StandardCard/index.js
+++ b/src/components/Cards/StandardCard/index.js
@@ -29,6 +29,7 @@ const ImageWrapper = styled.div`
   width: 100%;
 
   .no-image & {
+    background-color: transparent;
     display: flex;
     align-items: center;
     margin-bottom: ${spacing.xxsm};
--- a/src/components/Cards/shared/Sticker/index.js
+++ b/src/components/Cards/shared/Sticker/index.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import breakpoint from 'styled-components-breakpoint';
 import styled from 'styled-components';
 import { color, font, fontSize, letterSpacing, spacing } from '../../../../styles';
 import { Collection, VideoPlay } from '../../../DesignTokens/Icon';
@@ -28,6 +29,12 @@ const StyledSticker = styled.span`
     height: 1rem;
     max-height: 60%;
   }
+
+  ${breakpoint('xs', 'lg')`
+    &:not(:first-of-type) {
+      display: none;
+    }
+  `}
 `;
 
 const determineIconType = (contentType) => {


src/components/Cards/StandardCard/index.js
```diff
@@ -62,6 +63,7 @@ const StyledFavoriteButton = styled(FavoriteButton)`
 
 const StickerGroup = styled.div`
   display: flex;
+  flex-wrap: wrap;
```
meggles, line 66: I don't think stickers are supposed to wrap?

src/components/Cards/shared/Sticker/index.js
```diff
@@ -28,6 +28,11 @@ const StyledSticker = styled.span`
     height: 1rem;
     max-height: 60%;
   }
+
+  .no-image & {
+    margin-top: ${spacing.xxxsm};
+    margin-bottom: ${spacing.xxxsm};
```
meggles, line 34: 👍

src/components/Cards/StandardCard/index.js
jaslakson: Well, they have to do something...in mobile, just two strickers can cause it to wrap

meggles: Just confirmed this with Tabitha. They are always on one line.

jaslakson: Screen Shot 2020-04-09 at 4 50 57 PM

meggles: So, this may have fallen between the cracks, but there is a one-sticker max on mobile sized cards.

jaslakson: OK - I will update my code. I think it's probably best to hide it using css. @meggles do you agree?

meggles: I think so! The red stickers take priority over the black, so if we just hide the second sticker when there's two, that should work. `sticker + sticker { display: none; }`

jaslakson: OK - the only problem is that some stickers still overflow: 
Screen Shot 2020-04-09 at 5 03 26 PM

meggles: Do you mean the 'vegetarian' sticker? That's because there are two... there should always only be one sticker per card for widths below 1024px.

meggles: When I said 'mobile sized cards' I meant the smaller card size (there are 2 card sizes), which appears on widths below 1024px. That prob wasn't super clear on my part.

jaslakson: @meggles Updated to use have only one sticker on smaller cards. Thanks for your help!

PR #107, opened 2020-04-13
Tighten vertical space between filters

meggles:
This is based on some feedback from tabitha – site filters get a little more space than the others.


atk-dm:
:rocket: PR was released in `v0.1.10` :rocket:

PR #125, opened 2020-04-28
ARCHIE-559/576: search input issues found during QA

ttornese:


src/components/SearchInput/index.js
```diff
@@ -160,12 +172,12 @@ const StyledResetButton = styled.button`
...
 `;
 
-const ResetButton = connectCurrentRefinements(({ handleClick, items, refine }) => (
+const ResetButton = connectCurrentRefinements(({ handleClick, items, refine, query }) => (
   <StyledResetButton
     className="search-input__reset-button"
     type="reset"
     onClick={() => { if (handleClick) handleClick(); refine(items); }}
-    hidden={!items || !items.length}
+    hidden={!query || query.length === 0}
```
ttornese, line 180: reset button should be based on the value of query, not filters

PR #128, opened 2020-04-29
Fix sticker letterspacing, don't apply hover state on titles for touchscreens

meggles:


src/components/Cards/StandardCard/index.js
```diff
@@ -51,8 +51,10 @@ const TitleWrapper = styled.div`
 export const StyledTitle = styled(Title)`
   transition: color 0.2s ease;
 
-  &:hover {
-    color: ${color.mint};
+  @media(hover: hover) {
```
jaslakson, line 54: ooooohhhhh. nice!

jaslakson: are there wide spread changes we should be making for this type of thing? i.e. I would think this isn't the only place this would come in handy.

meggles: I think so! But... it's not supported by IE11 so we should be careful. I used it to for the favorites button hover state and also to position the 'done' button on the serp filter menu as an override – kind of a hacky use case, but it worked well?

PR #130, opened 2020-04-30
specify card link hover style

mirdathebirda:
Since there is no explicit hover state on the wrapper card link in Mise, Jarvis's base link styles are applied to the Card when hovered


atk-dm:
:rocket: PR was released in `v0.1.15` :rocket:

PR #136, opened 2020-05-06
Install and set up HTML addon

meggles:
The HTML addon prints out the compiled HTML for each component in a new tab in the addons panel.

Screen Shot 2020-05-06 at 12 52 10 PM


meggles:
> Any chance there's some sort of styled components addon that would remove the dynamically added class names?

I don't think so, unfortunately!

PR #138, opened 2020-05-11
SERP post release fixes

ttornese:


src/components/SearchBanner/index.js
```diff
@@ -28,23 +33,33 @@ const SearchBannerAnchor = styled.a`
...
 
 const SearchBannerClose = styled.button`
   height: 1.2rem;
   max-height: 1.2rem;
   max-width: 1.2rem;
   position: absolute;
-  right: 1.8rem;
-  top: 1.4rem;
+  right: 1rem;
+  top: 1rem;
```
meggles, line 50: If this to vertically center the close button, rather than setting the distance just from the top, it could be nicer to position it so it will be centered no matter what the height is (would also avoid having to override for tablet). you could do `top: 0; bottom: 0; margin-top: auto; margin-bottom: auto;`

src/components/SearchBanner/index.js
```diff
@@ -8,7 +9,7 @@ import { Close } from '../DesignTokens/Icon/svgs';
 
 const SearchBannerSection = styled.section`
   background-color: ${color.mint};
-  height: 4rem;
+  padding: 1.4rem 1.6rem;
```
meggles, line 12: could we use the spacing design token for 1.6rem here? `${spacing.sm}`

src/components/SearchBanner/index.js
```diff
@@ -28,23 +33,33 @@ const SearchBannerAnchor = styled.a`
   font: ${fontSize.md}/1 ${font.pnb};
   height: 100%;
   justify-content: center;
+  max-width: 28.5rem;
```
meggles, line 36: Isn't this just taking up 100% width of the parent container?

src/components/SearchBanner/index.js
```diff
@@ -19,6 +20,10 @@ const SearchBannerSection = styled.section`
       color: ${color.white};
     }
   }
+
+  ${breakpoint('tablet')`
+    height: 4rem;
```
meggles, line 25: Should this be adjusted using padding instead? Thinking about reusability, we should maybe try to avoid a situation where the content gets cut off.. or if this is specific to the search content, it might be better in jarvis..

PR #143, opened 2020-06-02
refactor algolia components and add tests

ttornese:


src/components/Algolia/shared/ResultsCount/ResultsCount.stories.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+import LabelFrame from '../../../LabelFrame';
+import ResultCount from './index';
```
meggles, line 4: I think the story & component should both be either `ResultCount` or `ResultsCount` (singular vs plural) – unless there's a reason for them to be different!

PR #149, opened 2020-06-08
resolve memory leak

jaslakson:
move `createGlobalStyle` into `jarvis` eliminates memory leak


atk-dm:
:rocket: PR was released in `v0.1.22` :rocket:

PR #150, opened 2020-06-10
setup global style for storybook

jaslakson:
do the `createGlobalStyle` thing for our storybook app.

Also, moving several of the dependencies to devDependencies so they don't get bundled with `jarvis` or `espresso`. Currently, we're getting build size errors on Heroku and our builds are taking forever because of the build size.


atk-dm:
:rocket: PR was released in `v0.1.23` :rocket:

PR #151, opened 2020-06-15
ARCHIE-726: create PersonHeadShot and PersonCard components

ttornese:


src/components/Person/PersonHeadShot/index.js
```diff
@@ -0,0 +1,35 @@
...
+);
+
+PersonHeadShot.propTypes = {
+  imgAlt: PropTypes.string,
+  imgSrc: PropTypes.string.isRequired,
+};
+
+
+PersonHeadShot.defaultProps = {
+  imgAlt: '',
```
ttornese, line 32: default to `''` for alt text. these headshots are presentational so screen readers can skip them

src/components/Person/PersonCard/PersonCard.stories.js
```diff
@@ -0,0 +1,16 @@
...
+import PersonCard from './index';
+
+export default {
+  title: 'Components|Person/PersonCard',
+  component: PersonCard,
+};
+
+export const Default = () => (
+  <PersonCard
+    description={`Bridget's favorite fair food is deep fried twinkies: learn why on this <a href="#">episode of proof</a>`} // eslint-disable-line
```
jaslakson, line 12: 💯

src/components/Person/PersonHeadShot/index.js
meggles: This is great – just want to check: does this actually render as an empty string in the html? I think I remember having to add a space in there, or else it just left the alt tag out altogether. But I could be wrong!

src/components/Person/PersonCard/index.js
```diff
@@ -0,0 +1,70 @@
...
+  height: 27.2rem;
+  justify-content: flex-start;
+  padding: 2rem 2.4rem 0;
+  width: 27.2rem;
+
+  .person-head-shot {
+    margin-bottom: ${spacing.md};
+  }
+
+  a {
```
meggles, line 22: I just thought of this – were you going to make this a class in the global styles so it can be reused elsewhere?

src/components/Person/PersonCard/index.js
```diff
@@ -0,0 +1,65 @@
...
+    ${mixins.styledLink(color.turquoise, color.darkerMint)}
```
meggles, line 23: 🎉

PR #152, opened 2020-06-15
Kids Search - Release 0.1.24

jaslakson:
* New icons for kids search
* new theme structure allowing extendable themes - similar to espresso, but with the added `light/dark` themes
* Updates to existing components to support kids search theme
* New basic refinement filter - without the expand/collapse


src/components/DesignTokens/Icon/svgs/Content.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+export default () => (
+  <svg xmlns="http://www.w3.org/2000/svg" width="18.152" height="21.162" viewBox="0 0 18.152 21.162">
```
meggles, line 4: Is this for decoration only? If so it needs `aria-hidden='true'`, `focusable='false'`, and `role='img'`?

src/components/DesignTokens/Icon/svgs/Sort.js
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export default () => (
+  <svg xmlns="http://www.w3.org/2000/svg" width="21.4" height="11.813" viewBox="0 0 21.4 11.813">
```
meggles, line 4: same comment as for the Content svg above.

atk-dm:
:rocket: PR was released in `v0.1.24` :rocket:

PR #153, opened 2020-06-15
links

mirdathebirda:
I'm not a huge fan of 'Styled Link' so I'm open to any other new names 😄

### DIFF
--- a/src/components/StyledLink/index.js
+++ b/src/components/StyledLink/index.js
@@ -1,0 +1,71 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/components/StyledLink/index.js
```diff
@@ -0,0 +1,71 @@
...
+        />
+      ) : null
+    }
+    {displayDecoration ? <span>⠇</span> : null }
+    {children}
+  </StyledLinkWrapper>
+);
+
+StyledLink.propTypes = {
+  children: PropTypes.node.isRequired,
```
mirdathebirda, line 59: Used `children` here instead of an `a` tag because I know some places in Espresso uses NextJS Links and I'm not sure if that's something we'd be passing through here or not

PR #155, opened 2020-06-17
Feature card

meggles:


src/components/Cards/FeatureCard/index.js
```diff
@@ -0,0 +1,184 @@
...
+  objectId,
+  onClick,
+  siteKey,
+  siteKeyFavorites,
+  target,
+  title,
+  href,
+}) {
+  return (
+    <StyledFeatureCard className={imageUrl ? '' : 'no-image'} contentType={contentType} data-testid="feature-card">
```
jaslakson, line 101: can you put these attributes on their own lines?

meggles: Sure!

PR #159, opened 2020-06-22
queue card

mirdathebirda:


src/components/Cards/QueueCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+import Title from '../shared/Title';
+import ProgressBar from '../shared/ProgressBar';
+
+const StyledQueueCard = styled.article`
+  position: relative;
+  width: 27.2rem;
+
+  h4 {
+    color: ${color.silver};
+    font: 12px/${lineHeight.sm} ${font.pnb};
```
meggles, line 18: I think 12px should be 1.2rem :)

src/components/Cards/QueueCard/index.js
```diff
@@ -0,0 +1,157 @@
...
+  width: 100%;
+`;
+
+export const StyledBadge = styled(Badge)`
+  position: absolute;
+  top: ${spacing.xsm};
+  left: ${spacing.xsm};
+
+  ${breakpoint('xs', 'lg')`
+    width: 1.6rem;
```
meggles, line 37: I actually don't think Badges on QueueCards size down for mobile.

src/components/Cards/QueueCard/index.js
mirdathebirda: oops good catch!

PR #161, opened 2020-06-22
ARCHIE-722: creating HeroCard component

ttornese:


src/components/DesignTokens/Logo/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Logos from './svgs';
+
+export const ATK = () => <Logos.ATK />;
+
+export const keyToLogo = key => ({
```
ttornese, line 6: more site & show logos will be added here as we get them

src/components/Cards/HeroCard/index.js
```diff
@@ -0,0 +1,184 @@
...
+
+const HeroCard = ({
+  backgroundImg, ctaHref, ctaText, description, iconKey, personHeadShot, sticker,
+}) => {
+  const Logo = keyToLogo(iconKey);
+  return (
+    <HeroCardWrapper backgroundImg={backgroundImg}>
+      <div>
+        {
+          Logo ? (
```
jaslakson, line 121: if the second half of the ternary is going to be null, it's more common to do 
```
Logo && (<something here />)
```

ttornese: i can update that 👍

src/components/Cards/HeroCard/index.js
```diff
@@ -0,0 +1,184 @@
...
+          sticker && (
+            <HeroCardSticker>
+              <Sticker {...sticker} />
+            </HeroCardSticker>
+          )
+        }
+        <HeroCardDescription cardType={personHeadShot && sticker ? 'learn' : 'watch'}>
+          {description}
+        </HeroCardDescription>
+        <HeroCardCta href={ctaHref}>
```
meggles, line 144: I think we've called these `ctaUrl` in other places – it might be good to keep it consistent?

ttornese: if we've been using `ctaUrl` it makes sense to keep using it! I'll change this over.

src/components/Cards/HeroCard/index.js
```diff
@@ -0,0 +1,184 @@
...
+    font: ${fontSize.lg}/1.22 ${font.pnb};
+    max-width: 24.6rem;
+  ` : `
+    font: ${fontSize.sm}/1.29 ${font.pnb};
+    max-width: 32.6rem;
+  `)}
+
+  ${breakpoint('md')`
+    font: ${fontSize.lg}/1.22 ${font.pnb};
+    max-width: 45.4rem;
```
meggles, line 82: Might want to check with design on these line heights to see if we can use the lineHeight.md token value (1.25) – the line height should probably be the same for the description. There's always going to be a slight discrepancy in the comps because XD doesn't allow them to specify unitless line height values, but then it gets converted to unitless in zeplin..

src/components/DesignTokens/Logo/Logo.stories.js
```diff
@@ -0,0 +1,12 @@
...
+import Logos from './index';
+
+export default {
+  title: 'Design Tokens|Logos',
+  component: Logos,
+};
+
+export const ATK = () => (
+  <Logos.ATK />
+);
```
meggles, line 12: 👏  hooray for adding logos

src/components/DesignTokens/Logo/svgs/ATK.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Atk = (props) => {
+  const { backgroundFill = '#D73A15', fill = '#FFFFFF' } = props;
```
meggles, line 5: Should we import and use the token names here?

src/components/Cards/HeroCard/index.js
ttornese: switched the description at all breakpoints to the md line height

src/components/DesignTokens/Logo/svgs/ATK.js
ttornese: done 👍

PR #163, opened 2020-06-23
Feature card CTA + center image for mobile

meggles:


src/components/Cards/FeatureCard/index.js
```diff
@@ -165,6 +178,11 @@ function FeatureCard({
           ) : null }
         </div>
       </a>
+      { ctaUrl ? (
```
jaslakson, line 181: could we do this here to be consistent?
```
{ ctaUrl && (
  <CtaLink ctaUrl={ctaUrl}>
    { ctaText }
  </CtaLink>
)
```

meggles: Sure, I can change that!

PR #164, opened 2020-06-23
Add Tall Card component

meggles:


src/components/Cards/TallCard/index.js
```diff
@@ -0,0 +1,202 @@
...
+
+  .tall-card__subcomponents-wrapper {
+    position: absolute;
+    bottom: 0;
+    left: 0;
+    padding: ${spacing.sm} ${spacing.xsm} ${spacing.xlg};
+    width: 100%;
+    display: flex;
+    flex-direction: column;
+    align-items: center;}
```
jaslakson, line 27: Something's off here - there's a `}`

meggles: whoops! thank you I will fix that

PR #165, opened 2020-06-23
ARCHIE-828: creating MediaObjectCard component

ttornese:


src/components/Cards/shared/PersonHeadShot/index.js
```diff
@@ -1,21 +1,40 @@
...
+      height: ${lg || md || sm}rem;
+      width: ${lg || md || sm}rem;
+    `}
+  `}
   overflow: hidden;
-  width: 10rem;
 `;
 
 const PersonHeadShotImg = styled.img`
   height: 100%;
```
meggles, line 28: If you add `object-fit: cover` on the image it will maintain the image aspect ratio while also centering the image and filling up the container.

src/components/Cards/shared/PersonHeadShot/index.js
```diff
@@ -1,21 +1,40 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
+import breakpoint from 'styled-components-breakpoint';
 
 const PersonHeadShotWrapper = styled.div`
-  border-radius: 10rem;
-  height: 10rem;
+  ${({ size: { sm, md, lg } }) => `
+    border-radius: ${sm}rem;
```
meggles, line 8: Not a huge deal, but if these are circle shapes it might be simpler to set border-radius to 50% once for all breakpoints.

src/components/Cards/MediaObjectCard/index.js
```diff
@@ -0,0 +1,93 @@
...
+import PersonHeadShot from '../shared/PersonHeadShot';
+import { color, font, fontSize, lineHeight, mixins, spacing } from '../../../styles';
+
+const MediaObjectCardWrapper = styled.section`
+  max-width: 100%;
+  width: 100%;
+
+  ${breakpoint('md')`
+    max-width: 27.2rem;
+    width: 27.2rem;
```
meggles, line 15: You could use ${grid.columnWidth} here if it makes sense to.

PR #168, opened 2020-06-29
ARCHIE-882: adding TextDecorations

ttornese:
edit: merged and approved here https://github.com/Americastestkitchen/mise-ui/pull/169


src/components/DesignTokens/TextDecoration/TextDecoration.stories.js
```diff
@@ -0,0 +1,44 @@
...
+
+import TextDecorations from './index';
+
+export default {
+  title: 'Design Tokens|TextDecorations',
+  component: TextDecorations,
+};
+
+export const CircleOne = () => (
+  <TextDecorations.CircleOne />
```
meggles, line 11: This looks great – would you mind adding the names underneath each like we do with icons? That way everyone can know which is which and what to call it.

src/styles/colors.js
```diff
@@ -10,6 +10,7 @@ const color = {
   darkerMint: '#5a8e91',
   turquoise: '#8fdee3',
   seaSalt: '#d1fcff',
+  seaMist: '#9DE1E6',
```
meggles, line 13: This looks _so close_ to the underlines for the person card links.. It might be worth asking if we need to introduce another color or if we can use the same for both?

src/components/DesignTokens/TextDecoration/index.js
```diff
@@ -0,0 +1,27 @@
...
+export const CircleTwo = () => <TextDecorations.CircleTwo />;
+export const CircleThree = () => <TextDecorations.CircleThree />;
+export const CircleFour = () => <TextDecorations.CircleFour />;
+export const CircleFive = () => <TextDecorations.CircleFive />;
+export const SquiggleOne = () => <TextDecorations.SquiggleOne />;
+export const SquiggleTwo = () => <TextDecorations.SquiggleTwo />;
+export const UnderlineOne = () => <TextDecorations.UnderlineOne />;
+export const UnderlineTwo = () => <TextDecorations.UnderlineTwo />;
+
+export const keyToLogo = key => ({
```
meggles, line 15: Would it maybe make more sense to call this `keyToGraphic` or `keyToTextDecoration` or something similar? Just since they aren't logos..

ttornese: yes, thats a typo leftover from me copying the helper from the Logo/index.js :)

src/styles/colors.js
ttornese: i'll check with design about that

PR #171, opened 2020-07-02
Audio Player

mirdathebirda:


src/components/AudioPlayer/AudioPlayer.stories.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+`;
+
+export const Default = () => {
+  /** retrieve information for podcast episode cards */
+  const [episodes, setEpisodes] = useState([]);
+  useEffect(() => {
+    async function fetchData() {
+      const response = await fetch(
+        'https://www-staging.americastestkitchen.com/api/v6/podcasts',
```
jaslakson, line 39: we should be sourcing this data from art19 directly since that's our goal for final implementation.

src/components/Cards/PodcastEpisodeCard/index.js
```diff
@@ -144,70 +149,131 @@ const TextWrapper = styled.div`
...
-        <>
-          <VideoPlay fill={color.white} />
-          <h3>{title}</h3>
-        </>
+class PodcastEpisodeCard extends Component {
+  componentDidMount() {
+    const { imageId } = this.props;
+    async function fetchData() {
+      const response = await fetch(
+        `https://art19.com/images/${imageId}`,
```
mirdathebirda, line 163: The image data below is gathered from making the same request through postman. Currently fetching data from this URL (through mise) will return an error -- we had a similar issue accessing data from their api before, so I'll follow up with Art19 about this.

src/components/AudioPlayer/AudioPlayer.stories.js
```diff
@@ -0,0 +1,101 @@
...
+  }
+`;
+
+export const Default = () => {
+  /** retrieve information for podcast episode cards */
+  const [episodes, setEpisodes] = useState([]);
+  useEffect(() => {
+    async function fetchData() {
+      const response = await fetch(
+        'https://art19.com/episodes?calendar_meta=true&page=1&per_page=10&rss=true&series_id=7f86ce08-696b-4741-a8fc-dbbd0cb3cee1',
```
jaslakson, line 40: This is good for the mise-ui story - let's build out a nice api lib for the espresso side.

PR #174, opened 2020-07-06
ARCHIE-898: adding play styles to Accordion, adding more stories for different Accordion use cases

ttornese:
* follow up PR will be done for play accordion content

### DIFF
--- a/src/components/Accordion/index.js
+++ b/src/components/Accordion/index.js
@@ -55,6 +55,28 @@ const AccordionButtonTheme = {
       width: 100%;
     `}
   `,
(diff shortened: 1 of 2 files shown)


src/components/Accordion/index.js
```diff
@@ -55,6 +55,28 @@ const AccordionButtonTheme = {
       width: 100%;
     `}
   `,
+  play: css`
```
meggles, line 58: I am thinking about whether we want themes that are so specific – maybe it would make more sense to name it something more reusable/generic? `dark`?

PR #175, opened 2020-07-07
Add white favorite button option to standard card

meggles:


src/components/Cards/StandardCard/StandardCard.stories.js
```diff
@@ -58,6 +58,7 @@ const ThemedComponent = ({
           displayCommentCount={displayCommentCount}
           displayFavoritesButton={displayFavoritesButton}
           displayLockIcon={displayLockIcon}
+          favoriteRibbonColor={`${color.white}`}
```
ttornese, line 61: i think you can drop the back tick string here and below because `color.white` returns a string

src/components/Cards/StandardCard/index.js
```diff
@@ -277,6 +280,7 @@ StandardCard.defaultProps = {
   displayCommentCount: false,
   displayFavoritesButton: false,
   displayLockIcon: false,
+  favoriteRibbonColor: `${color.eclipse}`,
```
mirdathebirda, line 283: I think you can drop the back ticks and curly braces `color.eclipse` here too

PR #176, opened 2020-07-08
move react scripts to devDependencies

jaslakson:
heroku complains about the 'slug' size on jarvis builds. The slug is basically the zip file that heroku gets from our repo's dependencies. `react-scripts` is huge and we don't need it in our main dependencies as far as I can tell. Moving it to `devDependencies` gets rid of the warning on jarvis and speeds up the jarvis builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@atk/mise-ui",
-  "version": "0.1.23",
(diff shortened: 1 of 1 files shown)


atk-dm:
:rocket: PR was released in `v0.1.25` :rocket:

PR #185, opened 2020-07-21
Add loading animation to loading card components

meggles:


src/styles/mixins.js
```diff
@@ -26,4 +28,40 @@ export default {
       }
     `;
   },
+
+  loadingGradientAnimation(
```
meggles, line 32: open to other names for this mixin if anyone has any ideas :)

PR #189, opened 2020-07-30
add now playing styles to podcast episode cards

mirdathebirda:


src/components/Cards/PodcastEpisodeCard/index.js
```diff
@@ -205,11 +220,12 @@ class PodcastEpisodeCard extends Component {
       imageId,
       siteKey,
       stickers,
+      isPlaying,
     } = this.props;
 
     return (
       <PodcastEpisodeCardWrapper
-        className="podcast-episode-card"
+        className={`podcast-episode-card ${isPlaying && 'is-playing'}`}
```
jaslakson, line 228: I think you'll want to change this to a ternary - if it's not playing, it might end up with a `false` classname

PR #190, opened 2020-08-04
ARCHIE-970 Newsletter signup on search announcement bar

douroudous:
- Handle new Algolia query rule params to indicate that a search banner should be shown to anonymous users only (`anonymousOnly`) and be tracked in Mixpanel (`mixpanelType`)
- If tracking banner in Mixpanel:
  * Create "Presented" event with `useEffect`
  * Add "Accepted" event to onClick
  * Both events refer to new `data-mixpanel-params` string

[Jarvis PR](https://github.com/Americastestkitchen/jarvis/pull/1316)

### DIFF
--- a/src/components/SearchBanner/index.js
+++ b/src/components/SearchBanner/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import breakpoint from 'styled-components-breakpoint';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@atk/mise-ui",
-  "version": "0.1.24",
+  "version": "0.1.26-canary.1a5e5c1b.0",
```
douroudous, line 3: will change after QA

atk-dm:
:rocket: PR was released in `v0.1.26` :rocket:

PR #192, opened 2020-08-10
Ja art19 proxy

jaslakson:
using proxy server url and multiple requests to art19 to finally get the image asset.

### DIFF
--- a/src/components/Cards/PodcastEpisodeCard/PodcastEpisodeCard.stories.js
+++ b/src/components/Cards/PodcastEpisodeCard/PodcastEpisodeCard.stories.js
@@ -1,7 +1,19 @@
 import React, { useEffect, useState } from 'react';
 import { withKnobs, text, select } from '@storybook/addon-knobs';
+import fetch from 'cross-fetch';
 
 import PodcastEpisodeCard from './index';
+
+const headers = {
+  method: 'GET',
+  headers: {
+    'Accept': 'application/vnd.api+json',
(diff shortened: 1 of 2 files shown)


src/components/Cards/PodcastEpisodeCard/PodcastEpisodeCard.stories.js
```diff
@@ -4,6 +4,17 @@ import fetch from 'cross-fetch';
 
 import PodcastEpisodeCard from './index';
 
+const headers = {
+  method: 'GET',
+  headers: {
+    'Accept': 'application/vnd.api+json',
+    'Authorization': 'Token token="jQ-ldsLx7USTQ4mjtOEBPIOsEXAno8UTh8l2KomNLQM5B75NUL-P9iFVGd1lF6c9-Lcq1dmUFTmhZTBsb09BmA", credential="14c39803-3e99-47c4-a4d3-d79398e74089"',
```
ccandelora, line 11: is there anyway to store these values in env variables?

src/components/Cards/PodcastEpisodeCard/PodcastEpisodeCard.stories.js
```diff
@@ -4,6 +4,17 @@ import fetch from 'cross-fetch';
...
+  },
+  mode: 'cors',
+};
+
+const proxyUrl = 'https://www-test.americastestkitchen.com/art19';
```
ccandelora, line 16: Not sure if you want to point to a local env here.

ccandelora: maybe make this an env variable too.

jaslakson: @ccandelora Good call - I updated as staging for now. I will look into the environment variable thing as well.

PR #194, opened 2020-08-11
Ja playlist poc

jaslakson:
Tweaks to Listable, Carousel-related styles and MediaObjectCard based on espresso integration work. The art19 header thing is secondary, and I think has already been accounted for elsewhere but including here for good measure.

### DIFF
--- a/src/components/AudioPlayer/AudioPlayer.stories.js
+++ b/src/components/AudioPlayer/AudioPlayer.stories.js
@@ -20,7 +20,7 @@ export default {
 
 const StyledLabelFrame = styled(LabelFrame)`
   padding: ${spacing.xlg} 0 ${spacing.sm};
-  
+
   h1 {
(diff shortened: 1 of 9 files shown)


src/components/AudioPlayer/AudioPlayer.stories.js
```diff
@@ -41,7 +41,7 @@ export const Default = () => {
         {
           method: 'GET',
           headers: {
-            'Accept': 'application/vnd.art19.v0+json;q=0.9,application/json;q=0.8',
+            'Accept': 'application/vnd.api+json',
```
mirdathebirda, line 44: I don't think we want to switch the accept header here. Using `application/vnd.api+json` doesn't give us a  response with all the information we're looking for.

jaslakson: @mirdathebirda I reverted this change. Thanks for catching that.

PR #196, opened 2020-08-12
adding test and dark mode styles to FilterButton

ttornese:


src/components/FilterButton/FilterButton.stories.js
```diff
@@ -1,13 +1,31 @@
 import React from 'react';
 import { action } from '@storybook/addon-actions';
+import styled, { ThemeProvider } from 'styled-components';
 
 import FilterButton from './index';
+import { color, spacing } from '../../styles';
+
+const DarkModeWrapper = styled.div`
+  background-color: ${color.nero};
```
octopushugs, line 9: I think we just changed the color name to `gunmetal`, right?

jaslakson: 🔫 🤘

PR #197, opened 2020-08-13
ARCHIE-1256 Search banner bug

douroudous:
- Add an default value (empty function) to `handleView` so that `useEffect` is run each time
- This avoids some Hooks-related errors that appeared to be due to running `useEffect` conditionally (most likely interfering with the [built in conditional effect firing](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect))

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@atk/mise-ui",
-  "version": "0.1.24",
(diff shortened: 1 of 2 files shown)


atk-dm:
:rocket: PR was released in `v0.1.27` :rocket:

PR #198, opened 2020-08-17
refactor refinement list to work with and without algolia

mirdathebirda:


jaslakson:
@mirdathebirda Looks like a test is failing

PR #201, opened 2020-08-20
email form with tests

jaslakson:
* Email form (just the input, button, error and how-we-use components
* Tests
* Fixing a warning about dependencies for the `useEffect` call in the audio component

This will be used in several ads/membership units.

### DIFF
--- a/src/components/Forms/EmailForm/index.js
+++ b/src/components/Forms/EmailForm/index.js
@@ -1,0 +1,218 @@
+import breakpoint from 'styled-components-breakpoint';
+import PropTypes from 'prop-types';
+import React, { useState } from 'react';
+import styled, { css } from 'styled-components';
+
+import {
+  color,
+  font,
+  fontSize,
+  lineHeight,
+  spacing,
+  withThemes,
+} from '../../../styles';
+
+import Button from '../../Buttons/Button';
+import FormInput from '../shared/FormInput';
+
+// eslint-disable-next-line no-useless-escape
+const emailRe = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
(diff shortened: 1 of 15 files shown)


src/components/Forms/EmailForm/index.js
```diff
@@ -0,0 +1,208 @@
...
+  return (
+    <EmailFormWrapper>
+      <EmailFormElement
+        buttonColor={buttonColor}
+        buttonTextColor={buttonTextColor}
+        disabled={disabled}
+        onSubmit={handleSubmit}
+      >
+        <FormInput
+          aria-label={inputLabel}
```
meggles, line 147: I think instead of the `aria-label` here, we should include a `label` element as part of this component that is visually hidden (or maybe with an option to expose it?). Better to use semantic html than aria.

src/components/Forms/EmailForm/index.js
```diff
@@ -0,0 +1,208 @@
...
+          error={error}
+          name="email"
+          placeholder={placeholder}
```
ttornese, line 150: im not totally sure how this will be parsed by a screen reader, but i remember reading that the `placeholder` attr doesnt work well with them. i think the typical pattern for an accessible input field would be to have a `<label>` with the place holder text, and a `for` attr with the `<input>`s `id` as its value. i do see there is an `aria-label` attr, so maybe that is enough for the screen reader

meggles: I agree! Also want to point out that often the placeholder is more of a prompt like "enter your work email" where the label might be more like "Email Address". Just mean that they can be different so we might want to keep both.

src/components/Forms/EmailForm/index.js
```diff
@@ -0,0 +1,208 @@
...
+        />
+        <Button
+          disabled={disabled}
+          type="submit"
+        >
+          {buttonText}
+        </Button>
+      </EmailFormElement>
+      <HowWeUseWrapper>
+        <HowWeUseLink
```
ttornese, line 161: for accessibility we may want to model this expand collapse behavior after the `Accordion` component in mise.

the the trigger button would have `aria-controls` attr equal to the id of the shown/hidden content, and `aria-expanded` as a boolean. 

the content would have an id to be use by the trigger button, and the `hidden` attr which could be set to the value of `showHow`.

PR #209, opened 2020-08-31
learn page changes

jaslakson:
* export text decorations for use in espresso
* cellAlign prop for carousels
* style favorites ribbons based on the `.favorited` classname
* test id for the loading card, so I can test in espresso

### DIFF
--- a/src/components/Cards/shared/FavoriteButton/index.js
+++ b/src/components/Cards/shared/FavoriteButton/index.js
@@ -1,54 +1,98 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
-import { color } from '../../../../styles';
+import styled, { css } from 'styled-components';
+import { color, withThemes } from '../../../../styles';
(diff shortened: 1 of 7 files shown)


src/components/Cards/shared/FavoriteButton/index.js
```diff
@@ -1,54 +1,98 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
-import { color } from '../../../../styles';
+import styled, { css } from 'styled-components';
+import { color, withThemes } from '../../../../styles';
 import { FavoriteRibbon } from '../../../DesignTokens/Icon';
 
-const StyledFavoriteButton = styled.button`
+const StyledFavoriteButtonTheme = {
```
jaslakson, line 7: This diff looks chunky, but the actual change isn't that involved

- Using themes
- adding styles for `.favorited`, because that's how the favorites widgets indicate an item is favorited

PR #211, opened 2020-08-31
single product book ad

jaslakson:
Single product ad for books


src/components/Ads/SingleProductAd/index.js
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    ${breakpoint('md')`
+      align-items: center;
+      display: flex;
+    `}
+  `,
+};
+
+const Product = styled.div.attrs({
```
meggles, line 39: Should we use `article` here too?

jaslakson: sure

PR #213, opened 2020-09-02
mostly lint fixes 

mirdathebirda:


src/components/AudioPlayer/index.js
```diff
@@ -235,7 +235,9 @@ const AudioPlayer = ({
           imageAlt={imageAlt}
           imageUrl={imageUrl}
         />
-        <span>•••</span><a href={href}>More From This Episode</a>
+        {
+          href && <><span>•••</span><a href={href}>More From This Episode</a></>
```
meggles, line 239: I think different screen readers handle reading bullets differently (some read them, some don't) – we might want to add an `aria-hidden` attribute on here to be safe.

PR #214, opened 2020-09-02
single membership ad, fix some naming issues and remove duplicate ad

jaslakson:
* we had two `MembershipShowCaseAd` components
* After talking to anne, changed the name of the `SingleProductMembershipAd` to `MembershipShowcaseAd` b/c that's what they've been calling it. I think we had our terminology mixed up.
* new `SingleMembershipAd` component
* `MediaObjectCardDescription` needs to accept html for the cast profiles
* Fix the exports of the circled text and underlined text items so they can be more easily accessed in `espresso`


src/components/Ads/SingleMembershipAd/__tests__/SingleMembershipAd.spec.js
```diff
@@ -0,0 +1,44 @@
...
+import SingleMembershipAd from '../index';
+import breakpoints from '../../../../styles/breakpoints';
+
+const defaultData = {
+  cta: 'do something',
+  ctaHref: '#foo',
+  title: () => <div>Cook with Confidence</div>,
+};
+
+describe('components', () => {
```
ttornese, line 14: can components be changed to something like `SingleMembershipAd should`

jaslakson: @ttornese Good call - I updated the tests for all Ad components to use the component name

PR #217, opened 2020-09-08
highlighted card

jaslakson:
Highlight card wraps Feature/Tall/Standard cards with a title and cta. This is similar to the document list carousel, but only ever designed for one card. This is featured on the video detail page, where the related recipe/episode/magazine are displayed.


ttornese (approved):

meggles:
Should we call this something different? `HighlightedCard` sounds to me like a new type of card, but this is really a wrapper for existing card types. I'm still thinking about what we'd call it though...

jaslakson:
@meggles I updated this PR with a different name for the component

meggles (approved):

PR #227, opened 2020-09-16
wrangling the linear-gradient

jaslakson:
* specific left and right positioning for card-carousel vs hero-carousel
* setup `gunmetal`, `jet` and `asphalt` flavas of gradients
* organize the `transparent` colors so I can find them

### DIFF
--- a/src/styles/colors.js
+++ b/src/styles/colors.js
@@ -1,8 +1,6 @@
 const color = {
   white: '#ffffff',
-  transparentWhite: 'rgba(255, 255, 255, 0.5)',
   black: '#000000',
-  transparentBlack: 'rgba(0, 0, 0, 0.7)',
   tomato: '#d73a15',
   rust: '#a72e12',
   whiteSmoke: '#f5f5f5',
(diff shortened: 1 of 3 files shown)


src/styles/colors.js
```diff
@@ -50,6 +47,11 @@ const color = {
   summerSky: '#3c9be2',
   mineShaft: '#2a2a2a',
   endeavour: '#0666B0',
+  transparentWhite: 'rgba(255, 255, 255, 0.5)',
+  transparentBlack: 'rgba(0, 0, 0, 0.7)',
+  transparentGunmetal: 'rgba(38, 38, 38, 0)',
```
meggles, line 52: Wouldn't this just be completely transparent with the alpha channel set to 0?

jaslakson: good catch. I think that's supposed to be 0.5

jaslakson: @meggles I updated the colors in the color file, but I also swapped over to transparent in the gradient to avoid any sort of 'starting line' on the leading edge of the gradient.

mirdathebirda (approved):

meggles (approved):

PR #228, opened 2020-09-17
create StationFinderForm component and add Arrow SVG

ttornese:
* also updated canary version so i dont have to link local espresso and mise anymore

### DIFF
--- a/src/components/DesignTokens/Icon/svgs/Arrow.js
+++ b/src/components/DesignTokens/Icon/svgs/Arrow.js
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { color } from '../../../../styles';
+
+const Arrow = ({ ariaHidden, fill, focusable }) => (
+  <svg
+    aria-hidden={ariaHidden}
+    focusable={focusable}
+    xmlns="http://www.w3.org/2000/svg"
+    viewBox="0 0 17.553 13.419"
+  >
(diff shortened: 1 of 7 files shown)


src/components/DesignTokens/Icon/svgs/Arrow.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { color } from '../../../../styles';
+
+const Arrow = ({ fill }) => (
+  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 17.553 13.419">
```
meggles, line 6: I think we'll need some of aria-attribute support on here – probably `aria-hidden` and `focusable=false` but would it make sense for these values be props so they can change based on the situation?

jaslakson (approved):

meggles (approved):

mirdathebirda (approved):

mirdathebirda (approved):

src/components/Forms/shared/FormInput/index.js
```diff
@@ -67,6 +68,7 @@ const FormTextInput = ({
         {label}
       </FormLabel>
       <FormInput
+        aria-label={ariaLabel}
```
meggles, line 71: How does this look in the DOM when its value is `null`? I just don't know if a screen reader would get tripped up on it. To be safe, maybe we should only include the attribute if it has a string value?

meggles (approved):

PR #244, opened 2020-10-01
stop player from being destroyed every time a new episode is selected

mirdathebirda:


src/components/AudioPlayer/index.js
```diff
@@ -185,6 +184,13 @@ const AudioPlayer = ({
     };
   }, [id, title]);
 
+  // destroy player and garbage collect
+  useEffect(() => {
+    if (player) {
+      return () => player.destroy();
+    }
+  }, []);
```
jaslakson, line 192: Looks like the build is failing b/c of a consistent return issue. You might just need to return true here when player is undefined.

jaslakson (approved):

meggles (approved):

ttornese (approved):

PR #258, opened 2020-10-09
manually remove aria-hidden on carousel cells when destroying carousel

jaslakson:
When flickity creates cells it adds aria-hidden and then selectively removes that attribute when cells are visible. When the instance is destroyed, there is code in Flickity source to remove the `aria-hidden` attributes. There's two problem right now: 

1) For this hero carousel, flickity isn't recognizing the three cards as visible so it's not automatically removing `aria-hidden` 
2) We destroy the instance b/c it's only one slide worth of cards but the `destroy` code in Flickity isn't removing the attributes - not sure why

So, when we destroy the flickity instance we should manually remove the `aria-hidden` from all cells.


meggles (approved):
Nice! 🎉

mirdathebirda (approved):

PR #267, opened 2020-10-16
Add transition to podcast episode card active & hover states

meggles:
Needed to separate the growing background container & image from the content of the card in order to apply a smooth transition animation.

### DIFF
--- a/src/components/Cards/PodcastEpisodeCard/index.js
+++ b/src/components/Cards/PodcastEpisodeCard/index.js
@@ -16,55 +16,72 @@ const PodcastEpisodeCardWrapper = styled.div`
   font: ${fontSize.md}/${lineHeight.md} ${font.pnr};
   padding: ${spacing.sm} ${spacing.md};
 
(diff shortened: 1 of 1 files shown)


ttornese (approved):

src/components/Cards/PodcastEpisodeCard/index.js
```diff
@@ -16,55 +16,72 @@ const PodcastEpisodeCardWrapper = styled.div`
...
+  }
+
   > div {
     display: flex;
-    overflow: hidden;
     max-height: 20rem;
   }
 
   p {
+    clear: both;
```
jaslakson, line 60: oooh - haven't seen one of these in a while! 😄

jaslakson (approved):
💯

src/components/Cards/PodcastEpisodeCard/index.js
meggles: Haha – sometimes you just need to use a float!

PR #271, opened 2020-10-19
make entire hero card clickable

jaslakson:
Make the entire hero card click-able


meggles (approved):
👍 Does this solve the `<a>` nested inside `<a>` error we were getting? It was showing up on a11y checks too.

ttornese (approved):

jaslakson:
@meggles Yes, this will solve that problem

PR #278, opened 2020-10-21
more podcast qa updates

mirdathebirda:
- make show more results button color #191919
- change title and description back to white
- add auto rotate option to carousels
- make entire podcast episode card clickable

### DIFF
--- a/src/components/Cards/PodcastEpisodeCard/index.js
+++ b/src/components/Cards/PodcastEpisodeCard/index.js
@@ -16,6 +16,10 @@ const PodcastEpisodeCardWrapper = styled.div`
   font: ${fontSize.md}/${lineHeight.md} ${font.pnr};
   padding: ${spacing.sm} ${spacing.md};
 
+  &:hover {
+    cursor: pointer;
+  }
+
   ${breakpoint('xs', 'md')`
     &.is-playing {
       background-color: ${color.jet};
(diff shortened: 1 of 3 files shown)


src/components/Carousels/Carousel/index.js
```diff
@@ -169,6 +169,7 @@ function temporaryCarouselFix() {
 function getFlickityInstance(el, options) {
   const {
     arrowShape,
+    autoPlay,
```
jaslakson, line 172: Does this need a default value?

src/components/Cards/PodcastEpisodeCard/index.js
```diff
@@ -252,6 +256,17 @@ class PodcastEpisodeCard extends Component {
...
+          episode,
+          title,
+          description,
+          href,
+          id,
+          imageAlt,
+          imageUrl,
+          siteKey,
+          stickers,
+        })}
```
meggles, line 269: I'm wondering if we need to put something on here for screen readers to indicate that this is interactive. It might be `role=button` but I'm not sure.. will look into it.

meggles: I think we should put `role="button" tabindex="0" aria-label="Play this episode"`

mirdathebirda: good catch! I'll add those changes now

meggles (approved):
💯❗

jaslakson (approved):

PR #284, opened 2020-10-22
only increase podcast episode card on larger screens

mirdathebirda:
the size increase of the selected podcast episode card made the entire podcast show page shift a littttle to the right on tablet portrait breakpoint


meggles (approved):
👍

jaslakson (approved):

PR #289, opened 2020-10-27
Make animation slower, animate scale instead of bg color

meggles:


jaslakson (approved):
On a *scale* of 1-10, this is a easily a 10.

mirdathebirda (approved):

PR #291, opened 2020-10-28
podcast episode cards should be set height and 84.4rem on large screens

mirdathebirda:


jaslakson (approved):

meggles (approved):

meggles:
Do we need to worry about the copy going long and getting cut off?

PR #293, opened 2020-10-29
Update button colors so they pass contrast ratio

meggles:
I decided to create new color names for these instead of updating the old names with new values to avoid ending up with color names with mismatched values across repos.


mirdathebirda (commented):
💯

mirdathebirda (approved):

jaslakson (approved):
🐸 👍

PR #295, opened 2020-10-30
update AudioPlayer image, aria-label for PodcastEpisodeCard and add hover state to carousel arrows

mirdathebirda:


src/components/AudioPlayer/index.js
```diff
@@ -263,11 +264,13 @@ const AudioPlayer = ({
...
       </PlyrPlayer>
       <LinkWrapper className="player__more-info">
-        <Image
-          aria-hidden="true"
-          imageAlt={imageAlt}
-          imageUrl={imageUrl}
-        />
+        <div className="player__image">
+          <Image
+            aria-hidden="true"
```
meggles, line 269: Should we remove this `aria-hidden` attribute? I think since it's an image, the alt tag will suffice.

jaslakson (approved):

meggles (approved):

PR #296, opened 2020-10-30
display fixes for school showcase ad

jaslakson:
Badge is included in the gif, so should be excluded form source. The cropping on the gif was causing issues so if we exclude gravity and aspect ratio we get the proper proportions.

### DIFF
--- a/src/components/Ads/ShowcaseAds/SchoolAd/index.js
+++ b/src/components/Ads/ShowcaseAds/SchoolAd/index.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import styled, { css } from 'styled-components';
 
-import Badge from '../../../Badge';
 import Gif from '../../../Gif';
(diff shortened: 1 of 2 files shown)


src/components/Ads/ShowcaseAds/SchoolAd/index.js
```diff
@@ -191,7 +184,7 @@ const deviceConfigMap = {
 const deviceIdMap = {
   desktop: 'mise-play/school-showcase-desktop-3',
   tablet: 'mise-play/school-showcase-tablet-3',
-  phone: 'mise-play/school-showcase-tablet-3',
+  phone: 'mise-play/school-showcase-desktop-3',
```
meggles, line 187: Is this right? should it be `school-showcase-phone-3`?

jaslakson: @meggles This is actually correct. There are two different files. phone and desktop get the same aspect ratio - I had it swapped before. Phone should be like desktop, not tablet.

PR #320, opened 2020-12-23
disable/enable flickty when switching between portrait and landscape mode on tablet

mirdathebirda:
the "carousel" gets stuck when you switch between tablet landscape and portrait since the number of slides changes between the breakpoints


octopushugs (approved):

src/components/Carousels/Carousel/index.js
```diff
@@ -252,6 +253,24 @@ const Carousel = ({ className, dotPosition, items, options, renderItem }) => {
     };
   }, [items, opts]);
 
+  useEffect(() => {
+    const handleResize = () => {
```
jaslakson, line 257: Can we throttle or debounce this? I know some browsers will fire a bunch of times as you resize.

mirdathebirda: 👍 good call!

PR #324, opened 2021-01-22
ARCHIE-1806 Article Card

mirdathebirda:


src/components/Cards/ArticleCard/index.js
```diff
@@ -0,0 +1,184 @@
...
+                key={text}
+                type={type}
+                text={text}
+              />
+            ))}
+          </StickerGroup>
+        ) : null }
+        <Title>{title}</Title>
+        <Description>{description}</Description>
+        <Byline>
```
jaslakson, line 149: I have heard mention of leveraging the "ByLine" in several places. Do you think it's worth separating this out into it's own component with tests, etc?

jaslakson (approved):
💯  for great tests - I left a question/suggestion to separate out the ByLine so that it could be leveraged in future projects, which can be done in a subsequent PR if you think it's a worth-while endeavor.

src/components/Cards/ArticleCard/index.js
mirdathebirda: good call! It looks like we'll be using it in Review Detail as well so I will separate it out in a following pr

ttornese (approved):

PR #325, opened 2021-01-25
ARCHIE-1830 allow flickity to determine height of carousels

mirdathebirda:
Allow Flickity to determine the height of the carousel rather than setting the heights ourselves. Vertical scrollbars will appear when the text of a card is longer than the set carousel height/card height.

Unloaded images have no size, which can throw off cell positions. The imagesLoaded option should re-positions cells once their images have loaded.


jaslakson (approved):
This is great - good find. This height stuff has been bothersome for a while.

ttornese (approved):

atk-dm:
:rocket: PR was released in `v1.2.0` :rocket:

PR #349, opened 2021-02-11
editors note and supporting icons

jaslakson:
After closer inspection of the comps, the `ReviewableSummaryItem` needed some tweaks. Plus, sometimes editors have revelations and when they do we want to be able adequately communicate those discoveries to our paying customers. To do so, we have crafted this handy-dandy `EditorsNote` component. What luck!


src/components/DesignTokens/Icon/svgs/Bell.js
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+
+function Icon() {
```
mirdathebirda, line 3: should this be bell?

ttornese (approved):

src/styles/colors.js
```diff
@@ -1,6 +1,7 @@
 const color = {
   white: '#ffffff',
   black: '#000000',
+  cuttySark: '#57797b',
```
mirdathebirda, line 4: i think cuttySark in espresso has a different hex value? it would be good to keep the color name/hex consistent

jaslakson: Good catch. Since this one is from mise, I'll adjust the espresso value.

mirdathebirda (approved):

PR #353, opened 2021-02-11
print styles, lazy image

jaslakson:
* Print styles for several elements
* Lazy load image by default


src/components/Accordion/index.js
```diff
@@ -151,6 +151,10 @@ const AccordionLabelWrapper = styled.div.attrs({
 
 const AccordionContent = styled.div`
   display: ${({ hidden }) => (hidden ? 'none' : 'block')};
+
+  @media print {
+    display: block !important;
```
jaslakson, line 156: The !important is required here because of the inline function above - for some reason that gets more priority than this print rule.

ttornese (approved):

mirdathebirda (approved):

PR #359, opened 2021-02-22
exclude plyr from local build

jaslakson:
externalize plyr from build - loading from cdn reduces the size of our JS build files. Since plyr is only used on a couple pages it makes sense to not included it everywhere.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -4,5 +4,11 @@
       "@babel/preset-env"
     ]
   ],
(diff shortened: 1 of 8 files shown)


ttornese (approved):

atk-dm (approved):

ccandelora (approved):

atk-dm:
:rocket: PR was released in `v1.4.0` :rocket:

PR #367, opened 2021-03-09
ARCHIE-2107 - load images when printing

jaslakson:
`useMedia` hook which lets us listen for change to print view...or breakpoints if we want.

Load images when we go into print mode using the custom dry event with media hook as fallback.

### DIFF
--- a/src/components/hooks/useMedia.js
+++ b/src/components/hooks/useMedia.js
@@ -1,0 +1,25 @@
+import { useEffect, useState } from 'react';
+
+const hasMedia = typeof window !== 'undefined' && window.matchMedia;
+
+function useMedia(query) {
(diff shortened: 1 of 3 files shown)


src/components/hooks/useMedia.js
```diff
@@ -0,0 +1,20 @@
+import { useEffect, useState } from 'react';
```
ttornese, line 1: cool!

ttornese (approved):

mirdathebirda (approved):

atk-dm:
:rocket: PR was released in [`v1.5.0`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.5.0) :rocket:

PR #383, opened 2021-03-29
CategoryCard Component for Landing-Bubbles + Tagline Links

jbonAA:


src/components/Cards/CategoryCard/CategoryCard.stories.js
```diff
@@ -0,0 +1,26 @@
...
+import CategoryCard from './index';
+
+export default {
+  title: 'Components|Cards/CategoryCard',
+  component: CategoryCard,
+};
+
+export const DefaultSvg = () => (
+    <CategoryCard 
+        assetType={'svgIcon'}
```
jaslakson, line 12: For string prop values we usually pass as double quotes - `assetType="svgIcon"`.

src/components/Cards/CategoryCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+    flex-direction: column;
+    height: 11.8rem;
+    justify-content: center;
+    width: 9.4rem;
+
+    ${breakpoint('xlg')`
+        &:hover {
+            border-radius: 8px;
+            box-shadow: 0 3px 6px 0 rgba(0, 0, 0, 0.16);
+            background-color: #ffffff;
```
jaslakson, line 22: even though this is just white, we typically refer to the colors using the variable name instead of the hex code.

src/components/Cards/CategoryCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+    background-color: ${color.frost};
+    border-radius: 50%;
+    display: flex;
+    height: 6rem;
+    justify-content: center;
+    margin-bottom: 10%;
+    position: relative;
+    width: 6rem;
+
+    ${props => props.iconId === 'shoppingCart' && css`
```
jaslakson, line 38: could we add a className to the shopping card svg and then have a scoped css rule here using that class name instead of the props function?

src/components/DesignTokens/Icon/svgs/ShoppingCart.js
```diff
@@ -0,0 +1,51 @@
...
+import { color } from '../../../../styles';
+
+const ShoppingCart = ({ fill }) => (
+    <svg 
+        xmlns="http://www.w3.org/2000/svg"
+        width="35.164" 
+        height="28.826"
+        fill={fill}
+    >
+        <g id="noun_Shopping_Cart_179069" transform="translate(-.6 -.3)">
```
jaslakson, line 12: let's drop all of these `id` attributes - same for the auto generated `className` attrs such as `cls-1`

src/components/DesignTokens/Icon/svgs/TrendingArrow.js
```diff
@@ -0,0 +1,27 @@
...
+import { color } from '../../../../styles';
+
+const TrendingArrow = ({ fill }) => (
+    <svg 
+        xmlns="http://www.w3.org/2000/svg" 
+        width="37.458" 
+        height="25.384"
+    >
+        <path 
+            data-name="Path 36217"
```
jaslakson, line 12: drop this attribute

jaslakson (changes requested):
This looks great overall - a couple of small requests - also, please double check the linter output. We usually only indent 2 spaces and I see lots of 4 in here. There's also a few other spacing issues around if/else, etc.

src/components/Cards/CategoryCard/__test__/CategoryCard.spec.js
```diff
@@ -0,0 +1,53 @@
...
+                svgId={'shoppingCart'}
+                tagline={'Browse All Reviews'}
+            />
+        )
+    }
+
+    const renderProductComponent = () => {
+        render(
+            <CategoryCard 
+                assetType={'productImage'}
```
ttornese, line 25: similar to jason's comment about these props can all be passed as with `"` and no `{`

src/components/Cards/CategoryCard/iconMap.js
```diff
@@ -0,0 +1,17 @@
...
+
+import ShoppingCart from '../../DesignTokens/Icon/svgs/ShoppingCart';
+import Star from '../../DesignTokens/Icon/svgs/Star';
+import TrendingArrow from '../../DesignTokens/Icon/svgs/TrendingArrow';
+import LandingPlay from '../../DesignTokens/Icon/svgs/LandingPlay';
+
+import { color } from '../../../styles'
+
+const IconMap = {
+    'shoppingCart': <ShoppingCart fill={color.mint}/>,
```
ttornese, line 11: with this syntax `(<ShoppingCart ... />)` the render functions for all of these icons will run every time this `IconMap` is called. i would try rewriting the map to look something like this:

```
const IconMap = {
  shoppingCart: ShoppingCart,
  ...
};
```

then inside the component that imports `IconMap`:

```
  const CategoryIcon = IconMap[svgId];
  ...
  return (
    <CategoryIcon fill={color.mint} />
  );
```

ttornese: it looks like the fill prop is the same color for every icon in the map so you probably don't need to include that in the map. it can be a dynamic prop passed to the `CategoryCard` or just hard coded in `CategoryCard`

src/components/DesignTokens/Icon/svgs/ShoppingCart.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { color } from '../../../../styles';
+
+const ShoppingCart = ({ fill }) => (
+    <svg 
+        xmlns="http://www.w3.org/2000/svg"
```
ttornese, line 7: these icons will be for display only. they dont add any extra context or functionality to the component. so for accessibility reasons they need should be hidden from screen readers and the tab index. you can add these attributes to the SVGs to make them a bit more accessible:

```
aria-hidden="true" // hide from screen reader
focusable="false" // remove from tab order
role="img" // given to svgs that act as images
```

src/components/Cards/CategoryCard/CategoryCard.stories.js
jbonAA: assetType="svgIcon"

src/components/Cards/CategoryCard/index.js
jbonAA: box-shadow: 0 3px 6px 0 ${color.white};

jbonAA: background-color: ${color.white};

src/components/Cards/CategoryCard/__test__/CategoryCard.spec.js
jbonAA: assetType="productImage"

src/components/Cards/CategoryCard/iconMap.js
jbonAA: Nice catch!

src/components/Cards/CategoryCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+        displayImage = (
+            <Image 
+                className={'category-product-image'}
+                imageAlt={tagline}
+                imageUrl={cloudinaryId}
+            />
+        );
+    }else{
+        displayImage = IconMap[svgId];
+    };
```
mirdathebirda, line 74: you can move this logic to line 81 and use a ternary and drop displayImage variable

src/components/Cards/CategoryCard/index.js
jbonAA: .shopping-cart-icon {
    position: absolute;
    left: 18%;
  }

src/components/Cards/CategoryCard/index.js
```diff
@@ -0,0 +1,99 @@
...
+        </LinkToBrowse>
+    );
+};
+
+
+CategoryCard.propTypes = {
+    assetType: PropTypes.oneOf(['productImage', 'svgIcon']).isRequired,
+    browsePath: PropTypes.string.isRequired,
+    cloudinaryId: PropTypes.string,
+    svgId: PropTypes.oneOf(['shoppingCart', 'star', 'trendingArrow', 'play', '']),
```
mirdathebirda, line 95: since these values aren't required, don't forget to add in some default props for them

src/components/Cards/CategoryCard/index.js
jbonAA:
```
const CategoryIcon = IconMap[svgId];

<ImageWrapper>
        {assetType === 'productImage' ? <Image className="category-product-image" imageAlt={tagline} imageUrl={cloudinaryId}/> : <CategoryIcon fill={color.mint} /> }
</ImageWrapper>
```

src/components/DesignTokens/Icon/svgs/LandingPlay.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { color } from '../../../../styles';
+
+const LandingPlay = ({ fill }) => (
+    <svg 
+        xmlns="http://www.w3.org/2000/svg"
+        width="30.001" 
+        height="30"
```
mirdathebirda, line 9: it would be good to use `viewbox` attribute instead of an explicit width and height for the svgs. viewbox preserves the aspect ratio which comes in handy when scaling

jbonAA:
```
viewBox="0 0 30.001 30"
```

src/components/Cards/CategoryCard/index.js
```diff
@@ -0,0 +1,88 @@
...
+`;
+
+const CategoryCard = ({
+  assetType,
+  browsePath,
+  cloudinaryId,
+  svgId,
+  tagline,
+}) => {
+  const CategoryIcon = IconMap[svgId];
```
jaslakson, line 59: @jbonAA I think you need to have either 1) a fallback if `svgId` doesn't exist in `IconMap` or 2) check to see if `svgId` is in `IconMap` before rendering, opting for `null` if it doesn't exist.

jbonAA: Opted for a fallback icon in the instance where svgId does not exist in IconMap

```
const CategoryIcon = IconMap?.[svgId] || IconMap['star'];
```

ttornese (approved):

jaslakson (approved):
💯

mirdathebirda (approved):
🔥 🎆 💯

PR #388, opened 2021-04-01
Category Card Updates + Adding CategoryCard to CardCarousel Types

jbonAA:
I added about a third of the category bubbles for testing in Storybook. 
Looking at it I'm realizing it's a lot... I can amend and reduce the number of entries for the stories significantly.

### DIFF
--- a/src/components/Cards/CategoryCard/index.js
+++ b/src/components/Cards/CategoryCard/index.js
@@ -1,11 +1,16 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import styled from 'styled-components';
 import breakpoint from 'styled-components-breakpoint';
+import PropTypes from 'prop-types';
+import React from 'react';
+import styled from 'styled-components';
 
 import IconMap from './iconMap';
 import Image from '../shared/Image';
 import { color, font, fontSize } from '../../../styles';
+import { getImageUrl } from '../../../lib/cloudinary';
+
+const CarouselContainer = styled.div`
+    height: 12.5rem;  
+`;
 
 const LinkToBrowse = styled.a`
   align-items: center;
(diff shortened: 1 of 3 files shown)


src/components/Cards/CategoryCard/index.js
```diff
@@ -74,24 +78,31 @@ const CategoryCard = ({
   svgId,
   tagline,
 }) => {
+  const [loading, setLoading] = useState(false);
```
jaslakson, line 81: What does this state value do? Looks like we are trying to avoid rendering on the server. Not sure what this is for.

src/components/Carousels/CardCarousel/index.js
```diff
@@ -52,14 +61,18 @@ const CardCarouselTheme = {
     }
 
     .carousel-cell {
-      margin-right: ${spacing.sm};
+      margin-right: ${({ type }) => (type === 'category' ? '.85rem' : `${spacing.sm}`)};
```
jaslakson, line 64: let's add the leading 0's to this as well.

jaslakson (commented):
I have a question about the useState in one of the components.

src/components/Cards/CategoryCard/index.js
jbonAA: The category bubble images weren't immediately rendering on the homepage until I scrolled through the carousel. I added this state value to delay the render momentarily so we have all the assets loaded.

src/components/Carousels/CardCarousel/index.js
jbonAA: I'll be mindful of that moving forward

src/components/Cards/CategoryCard/index.js
jbonAA: Reading through the Image component I used to display the cloudinaryIds I'm thinking I should've passed in the prop lazy: false to prevent the inlineSrc from rendering instead of the imageUrl

jbonAA: I updated the component to take in a bool for lazy and passed false for icons on the first view of the carousel

jaslakson (approved):

mirdathebirda (approved):

PR #415, opened 2021-04-21
accessibility updates for browse reviews page, specifically related to refinement list

ttornese:
While working on the accessibility QA card for the browse reviews page I ran into a ton of issues with `RefinementList` and `RefinementFilter` related to accessibility. Navigating through the filters with a screen reader made almost no sense. The readouts with these new components are much clearer.

I created two new temporary components (great names right?):
* `RefinementFilter2`
* `RefinementList2`

I also created a new wrapper for these components, `AccordionRefinementList`, which behaves similarly to the existing `SearchRefinementList` component.

These components are based off of the existing components, but built with a better foundation for accessibility. I did not want to edit the existing components as there are many pages across multiple brands tied to those components and it would have caused too much overhead. Moving forward as we have time, we can move each of those old pages over to these new refinement components.

Espesso implementation PR will follow.

### DIFF
--- a/src/components/Algolia/shared/ToggleRefinement/index.js
+++ b/src/components/Algolia/shared/ToggleRefinement/index.js
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connectRefinementList } from 'react-instantsearch-dom';
 
-import RefinementFilter from '../RefinementFilter/RefinementFilter';
+import RefinementFilter2 from '../RefinementFilter2';
 
 const ToggleRefinement = ({
   attribute,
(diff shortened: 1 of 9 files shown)


src/components/Algolia/shared/ToggleRefinement/index.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { connectRefinementList } from 'react-instantsearch-dom';
 
-import RefinementFilter from '../RefinementFilter/RefinementFilter';
+import RefinementFilter2 from '../RefinementFilter2';
```
ttornese, line 5: This component is only used on the new browse reviews page, so I changed it over to the new refinement filter.

jbonAA (approved):

jaslakson (approved):

mirdathebirda (approved):

atk-dm:
:rocket: PR was released in [`v1.6.0`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.6.0) :rocket:

PR #424, opened 2021-04-29
Finishing up reviews qa

jbonAA:


jaslakson (approved):

src/components/Carousels/DocumentListCarousel/index.js
```diff
@@ -157,6 +157,7 @@ const DocumentListCarousel = ({
       <Title>{title}</Title>
       {ctaText && ctaUrl && (
         <Cta
+          aria-label={ctaText}
```
mirdathebirda, line 160: you can omit this aria-label for this CTA since the text is listed below

jbonAA: This was one of those situations where I think the aria-label is needed. I just added it to prevent the text from reading the 'greater than' > when focused on the link

jbonAA: I added this aria-label to prevent the greater-than > being read out by screen reader after the ctaText

mirdathebirda (approved):

atk-dm:
:rocket: PR was released in [`v1.6.0`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.6.0) :rocket:

PR #432, opened 2021-05-03
ARCHIE-2321: create ArticleImage component for Articles

ttornese:
I also pulled the styled figcaption out of `PullQuote` and made it into the reusable `ArticleFigcaption` this is used for `PullQuote`, `ArticleImage` and will be used for other image components for this project.

### DIFF
--- a/src/components/Articles/ArticleImage/index.js
+++ b/src/components/Articles/ArticleImage/index.js
@@ -1,0 +1,115 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import breakpoint from 'styled-components-breakpoint';
+
+import ArticleFigcaption from '../shared/ArticleFigcaption';
+import { mixins } from '../../../styles';
+
+const ArticleImageFigure = styled.figure`
+  padding: 0;
+  margin: 0 0 2.8rem;
+
+  ${breakpoint('xs', 'xlg')`
+    .article-figcaption {
+      padding-bottom: 1.2rem;
+    }
+  `}
+
+  ${breakpoint('xlg')`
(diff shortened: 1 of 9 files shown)


src/components/Articles/ArticleImage/index.js
```diff
@@ -0,0 +1,103 @@
...
+  margin: 0 0 2.8rem;
+
+  ${breakpoint('xs', 'xlg')`
+    .article-figcaption {
+      padding-bottom: 1.2rem;
+    }
+  `}
+
+  ${breakpoint('xlg')`
+    ${({ width }) => {
```
jaslakson, line 21: would logic like this be handled easier if we did this by using classnames? something like `.width-default` and `.width-wide` or something like that. This just seems like a lot of logic for the styled component.

src/components/Articles/ArticleImage/index.js
```diff
@@ -0,0 +1,103 @@
...
+`;
+/* eslint-enable indent */
+
+const ArticleImagePicture = styled.picture`
+  img {
+    display: block;
+    margin-bottom: 1.2rem;
+
+    ${breakpoint('xlg')`
+      ${({ width }) => (`${mixins.articlesWidth(width)} ${width === 'default' ? 'margin-right: 1.6rem;' : ''}`)}
```
jaslakson, line 40: this would work well with a classname as well

jaslakson (commented):
Let me know what you think about using classnames instead of inline logic within the styled components.

jbonAA (approved):

mirdathebirda (approved):

jaslakson (approved):

PR #435, opened 2021-05-04
jb-ARCHIE-2344: Image List Component

jbonAA:
Waiting on confirmation regarding which html elements we should be accounting for / styling coming in from content.
Based off the following response from lines 374-464 in: 
https://github.com/Americastestkitchen/americano/blob/mp-sample-review-article/app/views/api/v6/review_sets/article_doc_sample.json

### DIFF
--- a/src/components/Articles/ImageList/ImageListContainer/ImageListContainer.stories.js
+++ b/src/components/Articles/ImageList/ImageListContainer/ImageListContainer.stories.js
@@ -1,0 +1,46 @@
+import React from 'react';
+
+import ImageListContainer from './index';
+import ImageListItem from '../ImageListItem/index';
+
+export default {
+    title: 'Components|Articles/ImageListContainer',
+    component: ImageListContainer,
+};
+
+const images = [{
+  altText: "picture of a thing",
+  cloudinaryId: "TnT/2020/1_CCJJ_Dill%20Pickles/SPS_Dill_Pickle_Hero_123",
+  content: "\u003cp\u003eThere are many things about this picture to say. There are many things about this picture to say. There are many things about this picture to say.\u003cp\u003e",
+  lazy: true
+}, {
+  altText: "picture of a thing",
+  cloudinaryId: "TnT/2020/1_CCJJ_Dill%20Pickles/SPS_Dill_Pickle_Hero_123",
+  content: "\u003cp\u003eThere are many things about this picture to say. There are many things about this picture to say. There are many things about this picture to say.\u003cp\u003e",
+  lazy: true
+}];
+
+const image = [{
+  altText: "picture of a thing",
+  cloudinaryId: "TnT/2020/1_CCJJ_Dill%20Pickles/SPS_Dill_Pickle_Hero_123",
+  content: "\u003cp\u003eThere are many things about this picture to say. There are many things about this picture to say. There are many things about this picture to say.\u003cp\u003e",
+  lazy: true
+}];
+
+export const defaultBoxOneImage = () => (
+  <ImageListContainer
+    children={image.map((el => <ImageListItem {...el}/>))}
+    className="image-list__container"
+    title="These are my images"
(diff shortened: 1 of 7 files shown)


src/components/ImageList/ImageListContainer/ImageListContainer.stories.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import ImageListContainer from './index';
+import ImageListItem from '../ImageListItem/index';
+
+export default {
+    title: 'Components|ImageList/ImageListContainer',
```
ttornese, line 7: could this be moved to `'Components|Articles/ImageListContainer`? Same with the ImageList item below.

src/components/ImageList/ImageListContainer/index.js
```diff
@@ -0,0 +1,58 @@
...
+    max-width: 69.6rem;
+    padding: 2.4rem;
+  `}
+
+  ${breakpoint('xlg')`
+    ${({ width }) => (mixins.articlesWidth(width))};
+  `}
+`;
+
+const ImageListTitle = styled.h2`
```
ttornese, line 25: I think this should be an `h3` all of the headings in the article section of the detail page will be nested under the `h2` "What We Learned"

src/components/ImageList/ImageListContainer/index.js
```diff
@@ -0,0 +1,58 @@
...
+);
+
+ImageListContainer.propTypes = {
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node,
+  ]).isRequired,
+  className: PropTypes.string,
+  title: PropTypes.string.isRequired,
+  width: PropTypes.oneOf(['default-box', 'wide-box']).isRequired,
```
ttornese, line 51: the width handler mixin is checking for `'default'` vs `'wide'` so i think the oneOf values here will need to be changed to that

mirdathebirda (approved):

src/components/ImageList/ImageListContainer/ImageListContainer.stories.js
```diff
@@ -0,0 +1,46 @@
...
+    children={image.map((el => <ImageListItem {...el}/>))}
+    className="image-list__container"
+    title="These are my images"
+    width="default"
+  />
+);
+
+export const wideBoxMultipleImages = () => (
+  <ImageListContainer
+    children={images.map((el => <ImageListItem {...el} width="wide-box"/>))}
```
jaslakson, line 41: This will need a `key` prop, otherwise you'll get some complaints from React.

src/components/ImageList/ImageListItem/index.js
```diff
@@ -0,0 +1,92 @@
...
+  color: ${color.eclipse};
+  font: ${fontSize.md}/2.4rem ${font.pnr};
+  width: 32.1rem;
+
+  ${breakpoint('md')`
+    width: 35.6rem;
+  `}    
+
+  ${breakpoint('xlg')`
+    ${({ width }) => (width === 'wide'
```
jaslakson, line 45: I like the idea of handling wide vs default using classnames vs logic embedded within the styled components. So, you would add `wide` as a classname and then add a style here. Tom had put together a few examples that might be good for reference.

src/components/ImageList/ImageListItem/index.js
```diff
@@ -0,0 +1,92 @@
...
+  altText,
+  cloudinaryId,
+  content,
+  deviceType,
+  lazy,
+  width,
+}) => (
+  <ListItemWrapper>
+    <Image
+      className="list-item__image"
```
jaslakson, line 67: Instead of relying on `deviceType` would it be better to use a Picture element with various srcsets based on screen width?

ttornese (approved):

PR #437, opened 2021-05-06
jb-ARCHIE-2322: Photo Collection Component

jbonAA:
Modeled after lines 103-143 for two or three images in collection: https://github.com/Americastestkitchen/americano/blob/mp-sample-review-article/app/views/api/v6/review_sets/article_doc_sample.json

### DIFF
--- a/src/components/Articles/ArticlePhotoCollection/index.js
+++ b/src/components/Articles/ArticlePhotoCollection/index.js
@@ -1,0 +1,195 @@
+import breakpoint from 'styled-components-breakpoint';
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+
+import ArticleFigcaption from '../shared/ArticleFigcaption';
+import { color, font, fontSize, mixins } from '../../../styles';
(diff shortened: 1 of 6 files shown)


src/components/Articles/ArticlePhotoCollection/index.js
```diff
@@ -0,0 +1,200 @@
...
+ArticlePhotoCollection.propTypes = {
+  caption: PropTypes.string,
+  count: PropTypes.oneOf([2, 3]).isRequired,
+  items: PropTypes.arrayOf(PropTypes.object).isRequired,
+  title: PropTypes.string,
+  width: PropTypes.oneOf(['default', 'wide']).isRequired,
+};
+
+ArticlePhotoCollection.defaultProps = {
+  caption: '',
```
ttornese, line 196: since these props are used in boolean checks in the render function i think their default value needs to be `null` or the boolean expression that uses them needs to be something like `caption.length > 0`

jaslakson (approved):

ttornese (approved):

PR #446, opened 2021-05-17
jb-ARCHIE-2454: Reviews Marketing Hat

jbonAA:


src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -0,0 +1,255 @@
...
+        <ContentSection isAnonymous={isAnonymous}>
+          <ContentWrapper isAnonymous={isAnonymous}>
+            <Headline>{headline}</Headline>
+            <Title>{title}</Title>
+            <Description isAnonymous={isAnonymous}>{description}</Description>
+            {isAnonymous ? (
+              <EmailForm
+                buttonText={buttonText}
+                inputId={inputId}
+                onSubmit={onSubmit.bind(this, subdomain)}
```
jaslakson, line 218: why do we need to do the `bind` here? Is there a use case were espresso or whoever is using this wouldn't have `subdomain` handy? We could also do this as a callback within this component if you wanted to pass it directly in as an argument.

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -0,0 +1,255 @@
...
+  subdomain,
+  title,
+  user,
+}) => (
+  <>
+    {user ? (
+      <MarketingHatWrapper>
+        <AdImage>
+          <source
+            srcSet={getImageUrl(desktopAsset, { aspectRatio: '16:9', width: 777, height: 307, crop: 'fill' })}
```
jaslakson, line 196: It might be interesting to try making `desktopAsset` and `mobileAsset` into prop values where the default value is what you have listed here. That would allow customization from the implementing codebase while having a good fallback. The other thing I notice here is that we typically define a helper method in `cloudinary.js` for each template type. I see that there's different heights/widths here so maybe you could combine the two approaches where you can centralize the aspectRatio and crop values but allow the width/height to be passed in as params.

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
jbonAA: Just following the proven logic from the article marketing hat component utilizing EmailForm. During implementation I can look for an alternative to the bind in espresso.

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
jbonAA: Updated the component w/ optional desktop and mobile asset props. Defaulting to match comp, but allows for customization from espresso. Additionally added reviewsMarketingHat template to cloudinary.js.

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -0,0 +1,258 @@
...
+  min-height: 4rem;
+  max-width: 100%;
+  text-transform: uppercase;
+
+  ${breakpoint('md')`
+    max-width: 22.6rem;
+  `}
+`;
+
+const Title = styled.h3`
```
ttornese, line 159: this `h3` needs to be nested under a relevant `h2`. i think maybe the styled component above this `Headline` should be an `h2`

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -0,0 +1,258 @@
...
+            <Description isAnonymous={isAnonymous}>{description}</Description>
+            {isAnonymous ? (
+              <EmailForm
+                buttonText={buttonText}
+                inputId={inputId}
+                onSubmit={onSubmit}
+                placeholder="Enter Your Email Address"
+              />
+            ) : (
+              <RegistrantSubmit
```
ttornese, line 221: I'm not sure about the behavior of this button but typically marketing hats send users to a different page once their associated button is clicked. If that's the behavior here this button element will need to be changed to an anchor tag with an href for accessibility purposes

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -0,0 +1,258 @@
...
+
+    .email-form__how {
+      left: 0;
+      padding-top: 0.7rem;
+      position: absolute;
+    }
+  `}
+`;
+
+const Description = styled.span`
```
ttornese, line 96: I think this should maybe be a p tag

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
jbonAA: Updated the button to an anchor tag w/ href, temporarily left it blank. Will connect and test during integration.

ttornese (approved):

jaslakson (approved):

PR #451, opened 2021-05-20
Updating Props for Ad Integration

jbonAA:
Updates:
- mdc and incode for hrefs: Mixpanel & order flows
- url and source key for carousel ad
- handling conditional render when !user in espresso

### DIFF
--- a/src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
+++ b/src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
@@ -182,57 +182,53 @@ const ReviewsMarketingHat = ({
   description,
   desktopAsset,
   headline,
+  incode,
   inputId,
   isAnonymous,
+  mdc,
   mobileAsset,
   onSubmit,
-  subdomain,
   title,
-  user,
 }) => (
-  <>
-    {user ? (
-      <MarketingHatWrapper>
-        <AdImage>
-          <source
-            srcSet={getImageUrl(desktopAsset, 'reviewsMarketingHat', 777, 307)}
-            media="(min-width: 1136px)"
+  <MarketingHatWrapper>
+    <AdImage>
+      <source
+        srcSet={getImageUrl(desktopAsset, 'reviewsMarketingHat', 777, 307)}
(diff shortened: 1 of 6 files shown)


src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -182,81 +182,76 @@ const ReviewsMarketingHat = ({
...
-    {user ? (
-      <MarketingHatWrapper>
-        <AdImage>
-          <source
-            srcSet={getImageUrl(desktopAsset, 'reviewsMarketingHat', 777, 307)}
-            media="(min-width: 1136px)"
+  <MarketingHatWrapper>
+    <AdImage>
+      <source
+        srcSet={getImageUrl(desktopAsset, 'reviewsMarketingHat', 777, 307)}
```
jaslakson, line 196: I am not sure this `getImageUrl` argument syntax is correct. If those last two arguments are width and height I think you need to make them an object:
```
{ width: 777, height: 307 }
```

or, am I wrong?

jbonAA:
```
const getImageUrl = (cloudinaryId, ...configArgs) => {
  if (!cloudinaryId) return null;
  const configKey = typeof configArgs[0] === 'string' ? configArgs.shift() : 'default';
  const key = imageConfig[configKey] ? configKey : 'default';
  const config = imageConfig[key].call(this, ...configArgs);
  if (config.aspectRatio) config.aspectRatio = normalizeAR(config.aspectRatio);
  return cloudinaryInstance.url(cloudinaryId, config);
};
```
I believe passing a string for the second argument of getImageUrl representing the imageConfig will shift the value, in this case 'reviewsMarketingHat' from configArgs and we'll be left with (777, 307) that are called as the params of this function: 

```
  reviewsMarketingHat(width, height) {
    return {
      aspectRatio: '16:9',
      crop: 'fill',
      height,
      width,
    };
  },
```

jaslakson (approved):

meggles (approved):

PR #452, opened 2021-05-24
jb-ARCHIE-2458: Landing Marketing Hat Integration

jbonAA:
Style & sizing updates for marketing hat

### DIFF
--- a/src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
+++ b/src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
@@ -6,12 +6,17 @@ import styled from 'styled-components';
 // grab email for for bottom logic
 import EmailForm from '../../../Forms/EmailForm';
 import { getImageUrl } from '../../../../lib/cloudinary';
-import { color, font, fontSize, letterSpacing } from '../../../../styles';
+import { color, font, fontSize, letterSpacing, spacing } from '../../../../styles';
 
 const AdImage = styled.picture`
   max-height: 26.7rem;
   width: 100%;
   z-index: 1;
+
+  img {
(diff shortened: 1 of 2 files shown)


src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -21,6 +26,16 @@ const AdImage = styled.picture`
...
+      min-width: 100%;
+    }
+  `}
+
+  ${breakpoint('xlg')`
+    min-width: 50%;
+
+    img {
+      min-height: 100%;
+      min-width: 100%;
```
meggles, line 38: Since you have a `min-width: 100%;` already set on `img`, you shouldn't need to include it here, or on line 29. Unless I'm missing something it's overriding?

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
```diff
@@ -35,17 +50,28 @@ const ContentSection = styled.div`
   width: calc(100% - 3.4rem);
   z-index: 2;
 
+  &.anon-user {
+    padding: 1.6rem 1.9rem 0.3rem 2rem;
```
meggles, line 54: Do any of these values align with the spacing design tokens we use in Mise?

meggles (approved):
Looks good other than the minor things I commented about!

src/components/Ads/ReviewsAds/ReviewsMarketingHat/index.js
jbonAA: Thanks! Good catch

jaslakson (approved):

PR #462, opened 2021-06-01
jb-ARCHIE-2559: Styles, Spacing, Triangle Bullet & New Phrasing

jbonAA:
- Update to email form props: optionalIcon

### DIFF
--- a/src/components/Forms/EmailForm/index.js
+++ b/src/components/Forms/EmailForm/index.js
@@ -136,6 +136,7 @@ const EmailForm = ({
   inputId,
   inputLabel,
   instanceId,
+  optionalIcon,
   placeholder,
   onSubmit,
 }) => {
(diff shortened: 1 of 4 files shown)


src/components/Forms/EmailForm/index.js
```diff
@@ -190,7 +191,7 @@ const EmailForm = ({
           role="button"
           type="submit"
         >
-          {buttonText}
+          {buttonText}<span>{optionalIcon}</span>
```
ttornese, line 194: does this need a null check since it is an optional prop?

jbonAA: Great call, I'll add that in now

jaslakson (approved):
agree with Tom, but 👍

ttornese (approved):

PR #475, opened 2021-06-22
Feature search focus

meggles:
Custom focus indicator for search page

### DIFF
--- a/src/components/Algolia/shared/RefinementFilter/RefinementFilter.js
+++ b/src/components/Algolia/shared/RefinementFilter/RefinementFilter.js
@@ -4,7 +4,7 @@ import styled, { css } from 'styled-components';
 
 import Badge from '../../../Badge';
 import { Checkmark } from '../../../DesignTokens/Icon/svgs';
-import { color, font, fontSize, spacing, withThemes } from '../../../../styles';
+import { color, font, fontSize, mixins, spacing, withThemes } from '../../../../styles';
 
 const RefinementFilterLabelTheme = {
   default: css`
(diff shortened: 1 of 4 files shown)


src/components/Algolia/shared/RefinementFilter/RefinementFilter.js
```diff
@@ -40,7 +40,7 @@ const RefinementFilterLabelTheme = {
     }
 
     &:focus-within {
-      border: 1px dashed ${color.eclipse};
+      outline: 2px dotted ${color.eclipse};
```
ttornese, line 43: What are your thoughts on making this focus outline treatment a mixin? Maybe with a color as an argument? We could also possibly import that mixin into espresso/jarvis from mise.

meggles: I think that's a good idea! I will add that in.

src/components/Algolia/shared/RefinementFilter/RefinementFilter.js
```diff
@@ -40,7 +40,8 @@ const RefinementFilterLabelTheme = {
     }
 
     &:focus-within {
-      border: 1px dashed ${color.eclipse};
+      ${mixins.focusIndicator(color.eclipse)};
+      outline-offset: 0;
```
meggles, line 44: Want to point out that I think `outline-offset` will usually be set to `3px`, but a couple of the algolia components look a little better without an offset so I have overridden it.

ttornese (approved):
🔥

mirdathebirda (approved):

src/styles/mixins.js
```diff
@@ -78,4 +78,11 @@ export default {
       }
     `;
   },
+
+  focusIndicator(color) {
```
jaslakson, line 82: would be be good to have a default here - whatever the most common situation would be. If we had to change it in the future we might be able to change fewer placements if we leverage a default.

meggles: good call – I will add a default.

jaslakson (approved):

PR #478, opened 2021-06-25
fixing styles for unordered lists inside of ArticleTextBlock

ttornese:


src/components/Articles/ArticleTextBlock/index.js
```diff
@@ -93,6 +93,30 @@ const ArticleTextBlockContent = styled.div`
     }
   }
 
+  ul {
```
jaslakson, line 96: do we have to worry about `ol` at all?

ttornese: it wasnt brought up by anne or put into the comps but i will double check with her

PR #500, opened 2021-07-22
atk_671: Geoff's Feedback Add aria-pressed to favorite ribbons

jbonAA:
- Change the aria-pressed value according to the visual state: "false" when the button is not selected; "true" when it is selected.

### DIFF
--- a/src/components/Cards/shared/FavoriteButton/index.js
+++ b/src/components/Cards/shared/FavoriteButton/index.js
@@ -105,6 +105,7 @@ const FavoriteButton = ({
 }) => (
   <StyledFavoriteButton
     aria-label={isFavorited ? `Remove ${title} from favorites` : `Save ${title} to favorites`}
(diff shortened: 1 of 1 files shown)


src/components/Cards/shared/FavoriteButton/index.js
```diff
@@ -105,6 +105,7 @@ const FavoriteButton = ({
 }) => (
   <StyledFavoriteButton
     aria-label={isFavorited ? `Remove ${title} from favorites` : `Save ${title} to favorites`}
+    aria-pressed={isFavorited ? 'true' : 'false'}
```
arindoneATK, line 108: I'll reach out via slack, but wondering if you need the ternary here if isFavorited is a boolean!

meggles (approved):

arindoneATK (approved):

jaslakson (approved):

PR #502, opened 2021-07-28
upgrading storybook to v6 for typescript configuration

arindoneATK:
This PR is to upgrade storybook. Meg and I are going to do some testing. I did leave typescript as a dev dependency, as well as some types. They shouldn't affect the upgrade at all.


jaslakson (approved):
looks fantastic - very exciting

meggles (approved):
🎉 🎉 🎉

ttornese (approved):

PR #503, opened 2021-07-28
jb-ARCHIE-2658

jbonAA:
If `buyNowOverrideAffiliateActive` is active and `buyNowOverrideAffiliateName` doesn't come from our list of affiliates don't default to amazon icon - only return button

### DIFF
--- a/src/components/Cards/ReviewableSummaryCard/index.js
+++ b/src/components/Cards/ReviewableSummaryCard/index.js
@@ -220,8 +220,9 @@ const ReviewableSummaryCard = React.memo(({
   const isDiscontinued = price?.toLowerCase()?.includes('discontinued') ?? false;
   const priceMarkup = price?.replace(parensRe, '<span>$1</span>') ?? null;
   let buyNowIcon = asin ? 'Amazon' : null;
-  if (buyNowOverrideAffiliateActive && buyNowOverrideAffiliateName) {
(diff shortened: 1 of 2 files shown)


src/components/Cards/ReviewableSummaryCard/index.js
```diff
@@ -220,8 +220,9 @@ const ReviewableSummaryCard = React.memo(({
   const isDiscontinued = price?.toLowerCase()?.includes('discontinued') ?? false;
   const priceMarkup = price?.replace(parensRe, '<span>$1</span>') ?? null;
   let buyNowIcon = asin ? 'Amazon' : null;
-  if (buyNowOverrideAffiliateActive && buyNowOverrideAffiliateName) {
-    buyNowIcon = buyNowOverrideAffiliateName;
+
+  if (buyNowOverrideAffiliateActive) {
```
arindoneATK, line 224: you may not need this outer conditional as if buyNowOverrideAffiliateActive is truthy, it will use that value, otherwise it will be set as null, so you can keep it a one liner. Let me know if I am wrong about that!

arindoneATK: Sorry, didn't realize the variable names were two different ones! I missed that one ended with Active and one ended with Name!

ttornese (approved):

PR #508, opened 2021-08-03
ARCHIE-2175: themeing PullQuote, adding control addon to storybook

ttornese:
Since Alex upgraded mise to Storybook 6 I decided to take a look at some new features/guidelines Storybook wrote for creating stories. I updated the `PullQuote` story to use these new guidelines. The doc pages I used to make updates were: [Args](https://storybook.js.org/docs/react/writing-stories/args), [ArgTypes](https://storybook.js.org/docs/react/api/argtypes), and [Controls](https://storybook.js.org/docs/react/essentials/controls).

Args map to props on a component, ArgTypes are objects you can use to define specific values and controls for Args, and Controls expose Args to users and let them be changed based on PropTypes or ArgTypes (similarly to knobs which will be replaced completely by Controls in Storybook 7). I think using controls could help use reduce the number of stories we have to write/maintain (ArticleTextBlock has a ton). It also gives QA/PO's and others easy an easy way to cycle through themes on components and also change certain configurations.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -9,5 +9,6 @@ module.exports = {
     '@storybook/addon-viewport/register',
     '@whitespace/storybook-addon-html',
     '@storybook/addon-notes/register',
+    '@storybook/addon-controls',
   ],
 };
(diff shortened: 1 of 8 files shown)


jaslakson (approved):

meggles (approved):
🎉

PR #509, opened 2021-08-05
Feature atk marketing hat

jbonAA:


src/components/MarketingHat/index.js
```diff
@@ -0,0 +1,315 @@
...
+
+  input {
+    background-color: ${color.white};
+    border-color: ${color.nobel};
+    font: ${fontSize.md}/1 ${font.mwr};
+    line-height: 4rem;
+    padding: 0 1rem;
+  }
+
+  #article-page-hat-form-submit {
```
jaslakson, line 64: What do you think about changing these id references to something more class-based? If we re-use this marketing hat elsewhere the `article-page` context of this id might not work as well. Plus, using id css rules feels like a sledge hammer - classnames always feels more natural.

ttornese (approved):

jaslakson (approved):

PR #514, opened 2021-08-17
suggestion card

jaslakson:
Borrowing the controls setup from @ttornese 

* Suggestion Card, will be wrapped by module to handle xhr, etc
* Haven't themed this for now b/c it will be in favorites, where there's only one style

### DIFF
--- a/src/components/Cards/SuggestionCard/__tests__/SuggestionCard.spec.js
+++ b/src/components/Cards/SuggestionCard/__tests__/SuggestionCard.spec.js
@@ -1,0 +1,85 @@
+import React from 'react';
+import { ThemeProvider } from 'styled-components';
+import { render, screen } from '@testing-library/react';
+import 'jest-styled-components';
+
+import SuggestionCard from '../index';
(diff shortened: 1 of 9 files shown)


ccandelora (approved):

harry-atk (approved):

src/components/Cards/SuggestionCard/__tests__/SuggestionCard.spec.js
```diff
@@ -0,0 +1,85 @@
...
+        <SuggestionCard
+          {...props}
+        />
+      </ThemeProvider>,
+    )
+  );
+
+  it('does renders an image', () => {
+    renderComponent(baseRecipe);
+    expect(screen.getByTestId('suggestion-img'));
```
ttornese, line 37: i think this test would pass even if an image wasn't passed to the component because the `suggestion-img` id is on the wrapper not the image itself. you could try the `getByAltText` query function to select the image itself

ttornese (approved):

PR #516, opened 2021-08-18
suggestion card carousel

jaslakson:
* Supplement `Carousel` with events and functionality for removing items and showing slides
* use `dek` vs `description` since that's what we get from the service
* change image orientation on suggestion cards - really close to having to go to a `srcset` approach to avoid cropping compliants
* whole suggestion card can't be a link b/c it has buttons in it - duh
* update `DocumentListCarousel` component stories...I started in this file, hoping to be able to use the component but breaking it out into `DocumengSuggestionCarousel` makes more sense.


harry-atk (approved):

src/components/Cards/SuggestionCard/index.js
```diff
@@ -24,33 +23,17 @@ const SuggestionCardLink = styled.a.attrs({
...
   ${breakpoint('md', 'lg')`
-    flex: 0 0 25rem;
-
-    img {
-      position: absolute;
-      margin-left: -50%;
-      max-height: 100%;
-      max-width: none;
-    }
+    flex: 0 0 25rem;;
```
harry-atk, line 36: quick lint fix

harry-atk (approved):
Looks good, quick lint fix on styles css. Its not something that should break css but I'm not sure how its parsed in styled css and what the tolerance is on that.

ttornese (approved):

PR #531, opened 2021-09-22
Typescript cheatsheet and pass on a few components

harry-atk:
I put some notes down that may be helpful. There's likely a lot of explanation missing or not necessary so I could use some feedback on what to add or remove.

Had these changes already going, so I will focus on cards, carousels, image next as discussed.

React.FC is not recommended by official sources, typing props and component is updated in docs. There are tradeoffs you want to remember. Some lint definitions rely on React.FC to enforce certain rules, some transforms like ones that convert typescript types to prop types for additional JS library support seem to also require explicitly defining a function as a react components. Anything that needs a solid identifier of a react functional component you should be suspect of this issue.

defaultProps is deprecated on functional components only. There's an issue with InferProps that would affect our ability to use es6 default parameters. I talked it over with Alex, and it seems best to use Typescript types directly on props instead of inferring them through prop types. We can leave the prop types there for js consumers like espresso and jarvis until espresso is migrated to typescript.  (jarvis has very little usage that shouldn't be a concern) There's seems to be some legacy in InferProps from flow that allows non required values to be null. This causes issues with defaultProps, as well as es6 default parameters, since defaults would need to be checked twice or with iconSize ??= 'default' in the body instead of the default parameters.


harry-atk:
Removed React.FC, recommended to type props and return type instead.

PR #536, opened 2021-09-28
ARCHIE-2840: CCO reviews design feedback

ttornese:


jbonAA (approved):

src/components/Ads/ReviewsAds/ReviewsEmailCapture/index.js
```diff
@@ -260,6 +262,7 @@ const AdWrapperTheme = {
 
     .email-form {
       .form-input input {
+        border: solid 1px #858585;
```
jaslakson, line 265: do we want names for these colors?

PR #539, opened 2021-10-05
ARCHIE-2931: creating new responsive TallToSquareCard for kids homepage

ttornese:


src/components/Cards/TallToSquareCard/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+  .${baseClassName}__sticker {
+    background-color: ${color.black};
+    margin: 0;
+    position: absolute;
+    opacity: 1;
+    top: -0.9rem;
+  }
+
+  ${breakpoint('md')`
```
jaslakson, line 34: It's strange that we call it a `desktop` image but we are using the `md` breakpoint. Do you think there would ever be a use-case for all three breakpoints (adding another image)?

jaslakson (approved):

jwinnfeild2517 (approved):

PR #563, opened 2021-11-29
ARCHIE-3039: Photo Carousel

harry-atk:
Jira: https://bostoncommonpress.atlassian.net/browse/ARCHIE-3039

@jaslakson should crossOrigin and decoding async be used here?


jaslakson (approved):
@harry-atk yes, please include those image attributes. I have a PR for mise-ui that will introduce those things elsewhere but definitely add it everywhere that you can.

jbonAA (approved):

PR #568, opened 2021-12-02
visual updates to SuggestionCard

twoodmanATK:


src/components/Cards/SuggestionCard/components/SuggestionCardDek.js
```diff
@@ -5,9 +5,13 @@ import { spacing } from '../../../../styles';
 const SuggestionCardDek = styled.div.attrs({
   className: 'suggestion-card__dek',
 })`
+  display: -webkit-box !important;
+  -webkit-line-clamp: 3;
+  -webkit-box-orient: vertical;
```
harry-atk, line 10: Mixin exists for this in styles/mixins/truncateLineClamp

src/components/Cards/SuggestionCard/components/SuggestionCardActions.js
```diff
@@ -14,12 +14,12 @@ const SuggestionCardActions = styled.div.attrs({
   }
 
   @media (min-width: 0) and (max-width: 413px) {
-    flex-direction: column;
+    flex-direction: column-reverse;
```
harry-atk, line 17: The reverse directions have accessibility issues. https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction#accessibility_concerns. I don't know if there is an easy fix for this.

harry-atk: You might have to use src/components/hooks/useMedia.js to change dom ordering.

twoodmanATK: 👍

twoodmanATK: Molly and I had a chat and we are just swapping the elements' order in the jsx, she agreed that it made sense for all breakpoints

ttornese (approved):

PR #596, opened 2022-01-06
ARCHIE-3188 add border to images list

harysvizcaino01:


harry-atk (changes requested):
the example in the comp seems to be an image list, but the recipe section on that page seems to be an article text block. Can we confirm which needs the box styling and when? Is it set in barista? This also needs to go in a withThemes({cco: css``}) so that it's only applied to cco

src/components/Articles/ArticleTextBlock/index.js
```diff
@@ -7,40 +7,42 @@ import ArticleTextBlockFloatImage from './components/ArticleTextBlockFloatImage'
...
 
-const ArticleTextBlockWrapper = styled.div`
-  margin-bottom: 2.4rem;
-  width: 100%;
-
-  &.article-text-block--box {
-    background-color: ${color.white};
-    padding: 2.4rem 1rem;
+const ArticleTextBlockWrapperTheme = {
+  default: css`
```
harry-atk, line 11: replace all css changes with:
${withThemes({
    cco: css`
      ${mixins.ccoReviewSetBorder()}
    `,
  })}

src/components/Articles/ArticleTextBlock/index.js
```diff
@@ -7,40 +7,42 @@ import ArticleTextBlockFloatImage from './components/ArticleTextBlockFloatImage'
...
+    margin-bottom: 2.4rem;
+    width: 100%;
+    border: ${({ borderWith }) => (borderWith ? `solid ${borderWith}px ${color.borderGray}` : '')};
+    padding: 5px;
```
harry-atk, line 15: padding here will probably break other designs for non border elements

src/components/Articles/ArticleTextBlock/index.js
```diff
@@ -215,6 +231,7 @@ const ArticleTextBlock = ({
   sidebarCard,
   title,
   width,
+  borderWith,
```
harry-atk, line 234: Prop removed with 
replace css changes with 
${withThemes({
    cco: css`
      ${mixins.ccoReviewSetBorder()}
    `,
  })}

harry-atk (changes requested):
Still missing the mark. This style falls under site theming and shouldn't become a prop requirement. Text block don't support conditional borders and the design provided is an image list which has border code already.

Please reach out to QA and Design to figure out what the requirements are here. If we do want borders to text blocks. Its probably only something like site === cco && displayOption = box

harry-atk:
If we need to change only articles and not review articles, we'll also need to be making this change on espresso with overrides instead of in mise

harry-atk:
fixed in espresso, going out in next qa build

PR #612, opened 2022-01-31
export ToggleRefinementMenu, update RefinementFilter2 logic

ttornese:


src/components/Algolia/shared/RefinementFilter2/index.js
```diff
@@ -139,39 +139,45 @@ const RefinementFilter = ({
...
-      e.preventDefault();
-      if (!isRefined && typeof handleClick === 'function') handleClick(e);
-      if (filterType === 'refinementList') {
-        refine(value);
-      } else if (filterType === 'toggleRefinement') {
-        if (currentRefinement.length > 0) {
-          refine(false);
-        } else {
+}) => {
+  let isActuallyRefined = isRefined;
```
ttornese, line 143: Multiple `ToggleRefinement` components can appear on a page with the same attribute, but different values. For example: `search_document_klass` is on reviews browse three times with the values `buying_guide`, `taste_test`, and `equipment_review`. Default behavior will refine all toggle refinements with the same filter. Overriding that behavior here so that only the `ToggleRefinement` with the appropriate value is actually refined.

jaslakson (approved):

PR #617, opened 2022-02-03
CORE-39

harry-atk:
refactors attributions in search card and adds to feature card
Going to cherry pick into a branch off of feature branch and implemented the below items.

### DIFF
--- a/src/components/Cards/StandardCard/index.js
+++ b/src/components/Cards/StandardCard/index.js
@@ -42,20 +42,6 @@ const StandardCardTheme = {
   `,
 };
 
-const RecipeAttribution = styled.div`
-  color: ${color.eclipse};
-  display: flex;
-  margin: 0.3rem 0 0.6rem 0;
-
-  .action-summary {
-    color: ${color.eclipse};
-  }
-
-  .icon--star {
-    margin-right: 1.6rem;
-  }
-`;
-
 const StyledStandardCard = styled.article`
   ${withThemes(StandardCardTheme)}
   margin-bottom: ${spacing.xsm};
--- a/src/components/Cards/StandardCard/index.js
+++ b/src/components/Cards/StandardCard/index.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import styled, { css, ThemeProvider } from 'styled-components';
 import breakpoint from 'styled-components-breakpoint';
 import { cards, color, fontSize, mixins, spacing, withThemes } from '../../../styles';
-import ActionSummaryItem from '../../ActionSummaryItem';
 import Attributions from '../shared/Attributions';
 import Badge from '../../Badge';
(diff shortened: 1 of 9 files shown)


src/components/Cards/shared/UserAttributions/UserAttributions.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <ActionSummaryItem icon="comment">
+        <strong aria-hidden="true">{commentsCount}</strong>
+      </ActionSummaryItem>
+      )}
+    </Wrapper>
+  );
+}
+
+export const RecipeUserAttributions = styled(UserAttributions)`
+  color: ${color.eclipse};
```
ttornese, line 58: might want to use your `cssThemedColor` mixin here and for the `.action-summary` text color. this text will need to match site themes once we render these cards on CIO and CCO. the SVG fills will need to match as well

ttornese: also not super important but we might want to rename this to something more generic like `StandardUserAttributions`. ratings will only appear on recipes but the comments can appear on any doc type with comments that is rendered by the `StandardCard`

src/components/Cards/StandardCard/index.js
```diff
@@ -275,28 +261,13 @@ function StandardCard({
...
-          )}
-          {searchComments && (
-            <ThemeProvider theme={{ siteKey: 'atk' }}>
-              <ActionSummaryItem icon="comment">
-                <strong aria-hidden="true">{searchComments}</strong>
-              </ActionSummaryItem>
-            </ThemeProvider>
-          )}
-        </RecipeAttribution>
+        <ThemeProvider theme={{ siteKey: 'atk' }}>
```
ttornese, line 264: I think this ThemeProvider should be dropped. Not sure why it was originally included but this will break the theming of this part of the card once it is used on non-atk site keys

src/components/Cards/shared/UserAttributions/UserAttributions.tsx
harry-atk: Yeah, I kept the name the wrapper it replaced was using which is definitely wrong.

src/components/Cards/StandardCard/index.js
harry-atk: I did carry it over for the existing implementation. For https://www.cookscountry.com/search would we be able to use the other site key colors here? the denim squirrel I believe

src/components/Cards/shared/UserAttributions/UserAttributions.tsx
harry-atk: It seems like the only usage of summary item is used in detail page header and is applying the theming already. It will be removed in new pr and I'll test if its there because of an override issue with espresso css

PR #620, opened 2022-02-07
ARCHIE-3166 - Topic carousel header

jaslakson:
New `TopicCarouselHeader` header bit for the basic carousel. Add in some tolerance for slight overflow when determining if all cells are visible. The right margin on the standard cards causes a slight overflow that Flickity picks up internally, making the pagination useless but still enabled because the `isAllSlidesVisible` check was not returning `true`.

### DIFF
--- a/src/components/Carousels/BaseCarousel/useFlickity.ts
+++ b/src/components/Carousels/BaseCarousel/useFlickity.ts
@@ -5,10 +5,16 @@ import React, { useCallback, useState, useEffect, useRef, MutableRefObject, Disp
 import useResizeObserver from 'use-resize-observer';
 import { fixIosScrollBehavior, fixLeftOverflow } from './bug-fixes';
 
+/**
+ * Allow for slight overflow related to margin on the last slide.
+ * @param flkty Flickity instance
+ * @returns Boolean if all slides are visible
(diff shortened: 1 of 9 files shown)


src/components/Carousels/BaseCarousel/useFlickity.ts
```diff
@@ -18,7 +24,6 @@ function isAllSlidesVisible(flkty: any): boolean {
  * @param flkty flickity instance
  */
 function disableFlickity(flkty: any): void {
-  flkty.options.wrapAround = false;
```
jaslakson, line 21: Note: this has been intentionally removed b/c it sets 'overflow: hidden' on the carousel slider and there seems to be a rendering bug where the first page of the carousel has a negative x transform with offset slides....so the slides look like they're in the right place but with the overflow off, nothing is visible.  This is related to the overflow tolerance change.

harry-atk (approved):

ttornese (approved):

atk-dm:
:rocket: PR was released in [`v1.33.4`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.33.4) :rocket:

PR #631, opened 2022-02-15
Text component with a background causes font styling issues

kayseriatk:


harry-atk (changes requested):
bold text here is provided by barista text editors. These wrap text in <strong> and <em> for bold and italic. I would change the font type to bold in those tags and not on all text components

harry-atk:
We probably want pnb for only bold right? not just bold that is also italic?

harry-atk (approved):

atk-dm:
:rocket: PR was released in [`v1.33.7`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.33.7) :rocket:

PR #692, opened 2022-04-12
CORE-299: styled byline list needs selectors to new text areas instead of color inheritance.

harry-atk:


harry-atk:
Let me actually export this from BylineList instead. Like we do with attributions. That way updates or usage to byline can see the light theme version

ttornese (approved):

harry-atk:
Going to cleanup the stories file here, was one of my first storybook files so it still has the deprecated knobs before I knew they were deprecated.

leo-fsl (approved):

PR #699, opened 2022-04-19
Jt hab 12

jwinnfeild2517:
1. Update the suggestion carousel to use the mise base carousel component
2. Updated suggestion card with favorites design


leo-fsl:
Some tests are failing, but it seems fine

leo-fsl (approved):

PR #713, opened 2022-05-02
Lp favorites dashboard mise

leo-fsl:


src/components/Cards/RelatedRecipeCard/RelatedRecipeCard.tsx
```diff
@@ -23,13 +23,15 @@ const ContentTheme = {
     background: ${color.white};
     display: flex;
     flex-direction: column;
+    width: 21.2rem;
     max-width: 21.2rem;
```
jwinnfeild2517, line 27: add height and max-height here so the card does not increase in size since the length of the titles will be different.

leo-fsl: I don't think will be good to set a height here, because it can change if the card have the avgRating or not.

So it can change depending on how we use the card on espresso. But the parts of the component itself handle its on height.

Did you see any problems with the height?

jwinnfeild2517 (approved):

PR #715, opened 2022-05-03
Favorites hover color

leo-fsl:


src/components/Cards/shared/Title/index.js
```diff
@@ -10,7 +10,7 @@ const TitleTheme = {
 
     @media(hover: hover) {
       &:hover {
-        color: ${color.mint};
+        color: ${color.grayishCyan};
```
jwinnfeild2517, line 13: I would not change the shared component here as it will also be changed in other areas of the site.

jwinnfeild2517 (commented):
The madeForyou card also needs an HoverState According to the card

src/components/Cards/shared/Title/index.js
leo-fsl: Removed change

src/components/Cards/MadeForYouCard/index.js
```diff
@@ -16,6 +16,12 @@ const MadeForYouCardWrapper = styled.a`
     height: 230px;
     display: block;
   }
+
+  &:hover {
+    h3 {
+      color: ${color.grayishCyan};
```
jwinnfeild2517, line 22: I think the entire card also has to shift up a bit

src/components/Cards/RelatedRecipeCard/RelatedRecipeCard.tsx
```diff
@@ -58,6 +53,10 @@ const Headline = styled.span`
   ${withThemes(HeadlineTheme)}
   ${cssThemedColor}
   ${mixins.truncateLineClamp(3)}
+
```
jwinnfeild2517, line 56: place the hover state on the entire card. So when the user hovers on the card. the title changes color

src/components/Cards/MadeForYouCard/index.js
leo-fsl: Done

src/components/Cards/RelatedRecipeCard/RelatedRecipeCard.tsx
leo-fsl: As said, the hover is only on the title

PR #723, opened 2022-05-10
CORE-370: ad belt design fixes

ttornese:


src/components/Ads/ReviewsAds/ReviewsEmailCapture/index.js
```diff
@@ -228,8 +235,9 @@ const AdWrapper = styled.div`
 
     .email-form-wrapper {
       ${({ isWide }) => (isWide ? `
-        max-width: calc(50% - 3rem);
-        width: calc(50% - 3rem);
+        height: 4rem;
+        max-width: calc(55% - 3rem);
+        width: calc(5560% - 3rem);
```
harry-atk, line 240: 5560% wouldn't affect with max width but typo right

harry-atk (changes requested):
Could we keep the side effect styles out in https://github.com/Americastestkitchen/espresso/pull/3160/files and just swap the styled elements out with extended ones with just those values. Text trial belt should only be on homepage work so far.

```
export const VariantButtonArea = styled(ButtonArea)`
  max-width: 100%;
`;

export const VariantDescription = styled(Description)`
  max-width: 27.5rem;
  ${md(css`max-width: 42rem;`)}
`;
```

harry-atk (approved):

PR #736, opened 2022-05-17
content-94 Update to Related Content Card in Mise

tdorin:
A featured content component without an image is showing a blank image. The content should shift to the left to adjust for the lack of image. Updated the knobs to use controls instead within Storybook.


harry-atk (approved):

ttornese (approved):
🎊

PR #741, opened 2022-05-18
CORE-405: re-order FeatureCard children to update reading order of the card when using a screen reader

ttornese:


src/components/Cards/FeatureCard/index.js
```diff
@@ -249,6 +238,17 @@ function FeatureCard({
             </div>
           )}
         </div>
+        <StyledBadge className={className} type={siteKey} />
```
ttornese, line 241: These elements are positioned absolutely, so card layout will remain the same. Right now the badge and favorite button are being read by screen readers before the document title and doc type. Moving these children further down will make the reading order more appropriate.

PR #774, opened 2022-06-06
Mise Badges and Storybook

tdorin:
updated Badge component in Mise to a .ts file and removed the `knobs` and using `controls` now


.eslintrc
```diff
@@ -10,10 +10,15 @@
...
   },
+  "env": {
+    "browser": true,
+    "node": true
+  },
   "parser": "@typescript-eslint/parser",
   "rules": {
     // Soft some rules.
     "no-use-before-define": "off",
+    "@typescript-eslint/no-explicit-any": "off",
```
harry-atk, line 21: We turned no explicit any off in espresso only because there's a lot of untyped api calls, but we really should be able to stay away from explicit any in mise and hopefully can add this rule back to espresso in the future. I can look through this pr and suggest types in a minute or two

harry-atk (changes requested):

.eslintrc
tdorin: That would be great!

harry-atk:
@tdorin I'm actually thinking explicit any is fine for now. We can always go back and clean up later. It seems like though that there are existing issues present with feature_v2 and the dependency updates. I would maybe consider pointing to develop if we are making implementation changes now or in the future.

PR #779, opened 2022-06-09
FEA-25: converting PullQuote component and stories to typescript

ttornese:
Converting `PullQuote` article component and its stories to typescript.

Created reusable type `ArticleComponentWidthType` which all article components can pull in as we convert them to typescript. Thoughts on the folder location and whether we want to move it somewhere else?


bradwaropay:
If that is truly just an article relevant type I think that is a good pattern. Eventually we won't need types for any CSS in JS but that could be useful for shared types across component variations.

bradwaropay (approved):

PR #781, opened 2022-06-13
CORE-480: atk homepage a11y screen reader fixes

ttornese:
* re-order children of ArticleCard so screen reader reads them in the correct order
* hide decoration character from screen reader on form signup button

### DIFF
--- a/src/components/Cards/ArticleCard/ArticleCard.tsx
+++ b/src/components/Cards/ArticleCard/ArticleCard.tsx
@@ -71,6 +71,7 @@ const CardBody = styled.div`
   justify-content: center;
   gap: 8px;
   padding: 16px;
+  order: 1;
 `;
 
 const Title = styled.h3`
@@ -126,6 +127,9 @@ function SplitCard({
 
   return (
     <Card data-qa="article-card" {...linkProps}>
+      <CardBody>
+        {children}
+      </CardBody>
       {!imageError ? (
         <Stack>
           <CardImage onError={() => setImageError(true)}>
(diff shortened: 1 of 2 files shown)


src/components/Cards/ArticleCard/ArticleCard.tsx
```diff
@@ -126,6 +127,9 @@ function SplitCard({
 
   return (
     <Card data-qa="article-card" {...linkProps}>
+      <CardBody>
```
ttornese, line 130: Ordering was changed here so document information is read by a screen reader before the badge & favorite information.

src/components/Cards/ArticleCard/ArticleCard.tsx
```diff
@@ -71,6 +71,7 @@ const CardBody = styled.div`
   justify-content: center;
   gap: 8px;
   padding: 16px;
+  order: 1;
```
oddwili, line 74: If the template was updated to have this component first, is this line necessary?

oddwili (changes requested):
small question

src/components/Cards/ArticleCard/ArticleCard.tsx
ttornese: Yes, it's a bit of a strange case. I needed to re-order the template so the screen reader reads the `CardBody` component's content to a user first, before the content of the `Stack` component. I added the order style here because visually, the `CardBody` component needs to be displayed below the `Stack` component (mobile) or to the right (tablet and above).

oddwili (approved):

atk-dm:
:rocket: PR was released in [`v1.33.24`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.33.24) :rocket:

PR #787, opened 2022-06-14
refactor: CardWrapper and associated Card components typescript

oddwili:
### What does this PR do?
adds types for CardWrapper, StandardCard, FeatureCard, and TallCard. Creates a BaseCard type that each component extends from.

Note: TS is throwing a lot of errors, but this was scope was getting bigger and bigger so I'm ignoring them for now and will address in another PR.

### How should I review this PR?
Read through the changes and let me know what you think! Do you like having the `Cards.d.types`? Or should we have a specific `types/` directory?


src/components/Cards/LeadMarqueeCard/LeadMarqueeCard.tsx
```diff
@@ -238,7 +238,7 @@ type LeadMarqueeCardProps = {
   /** Image for card. */
   imageUrl: string;
   href: string;
-  siteKey: 'atk' | 'cco' | 'cio' | 'kids' | 'school' | 'shop';
+  siteKey: DomainSiteKey | 'kids' | 'school' | 'shop';
```
tdorin, line 241: what about using the `ThemeSiteKey` here that you created in the `index.d.ts`?

src/components/AudioPlayer/index.js
```diff
@@ -200,12 +200,14 @@ const AudioPlayer = ({
         ],
       };
     } else {
+      // eslint-disable-next-line no-undef
```
tdorin, line 203: No lie! I was having the same errors on the `setTimeout` and `clearTimeout` as well!
I added this to my `.eslint` file and it cleared the lint errors so these comments could be removed.
```
"env": {
    "browser": true,
    "node": true
  },
  ```
   But as of this morning, I removed it, and those lint errors are no more. so, I can only assume it's goblins.

src/components/Cards/LeadMarqueeCard/LeadMarqueeCard.tsx
oddwili: I needed to re-base! it should be good now

oddwili:
Closing because I pulled changes into smaller PRs and merged in.

PR #795, opened 2022-06-16
surveycards typescript conversion

jwinnfeild2517:
Converged SurveyCards into typScript


src/components/Cards/SurveyCard/SurveyCard.stories.tsx
```diff
@@ -10,7 +10,10 @@ export default {
   decorators: [withKnobs],
 };
 
-const handleSubmit = () => {};
+const handleSubmit = () => {
+  // eslint-disable-next-line no-console
```
bradwaropay, line 14: Not sure I follow what we are trying to do with this handle submit. If this is just to fire off a console log we should get rid of it.

bradwaropay (changes requested):

oddwili (approved):

PR #800, opened 2022-06-22
converge made for you card to typescript

jwinnfeild2517:


oddwili (changes requested):
some small things, otherwise looks good

PR #804, opened 2022-06-29
suggestion card tsx

jwinnfeild2517:
Suggestion Card and Document Suggestion Carousel typeScript Conversion


src/components/Cards/shared/UserAttributions/UserAttributions.tsx
```diff
@@ -17,8 +17,8 @@ const Wrapper = styled.div`
 `;
 
 export type UserAttributionsProps = InferStyledTypes<typeof Wrapper> & {
-  avgRating?: number;
-  numRatings?: number;
+  avgRating?: number | null;
+  numRatings?: number | null;
```
oddwili, line 21: if optional, could we just keep this to `number`?

src/components/DesignTokens/Icon/svgs/Save.tsx
```diff
@@ -1,8 +1,13 @@
...
 
-const Save = ({ fill }) => (
+type SaveRibbonTypes = {
+  fill: string,
+  className?: string,
+}
+
+const Save = ({ fill, className }: SaveRibbonTypes) => (
   <svg
+    className={className}
```
oddwili, line 10: if this is not passed in it will set `undefined` as the classname

oddwili (changes requested):

src/components/Cards/shared/UserAttributions/UserAttributions.tsx
jwinnfeild2517: The current API response returns null when these values are not available.

oddwili (approved):

PR #827, opened 2022-07-13
personcard to tsx

jwinnfeild2517:
## What does this PR do

Convert person card to typeScript

### DIFF
--- a/src/components/Cards/PersonCard/index.tsx
+++ b/src/components/Cards/PersonCard/index.tsx
@@ -4,9 +4,7 @@ import styled, { css } from 'styled-components';
 import PersonHeadShot from '../shared/PersonHeadShot/PersonHeadShot';
 import { color, font, fontSize, lineHeight, mixins, spacing, withThemes } from '../../../styles';
 
-type ModeType = {
-  mode: string
-}
+type ModeType = 'dark' | 'light';
 
 const PersonCardWrapperTheme = {
   default: css`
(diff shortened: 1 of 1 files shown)


src/components/Cards/PersonCard/index.tsx
```diff
@@ -1,10 +1,13 @@
 import React from 'react';
-import PropTypes from 'prop-types';
 import styled, { css } from 'styled-components';
 
 import PersonHeadShot from '../shared/PersonHeadShot/PersonHeadShot';
 import { color, font, fontSize, lineHeight, mixins, spacing, withThemes } from '../../../styles';
 
+type ModeType = {
+  mode: string
+}
```
ttornese, line 9: we only have dark and light mode, so a better type here might be something like: `type ModeType = 'dark' | 'light'`. this is a good candidate for a global type as well

src/components/Cards/PersonCard/index.tsx
```diff
@@ -70,11 +73,25 @@ const PersonCardDescriptionTheme = {
...
+const PersonCardDescription = styled.div<ModeType>`
   ${withThemes(PersonCardDescriptionTheme)}
 `;
 
-const PersonCard = ({ description, imgAlt, imgCloudinaryId, mode, name }) => (
+type PersonCardProps = {
+  description: string,
+  imgAlt?: string,
+  imgCloudinaryId: string,
+  mode?: string,
```
ttornese, line 84: if you redefine `ModeType` to what i made in my comment above you could then set the type of this `mode` prop to `ModeType`

src/components/Cards/PersonCard/index.tsx
jwinnfeild2517: 👍

src/components/Cards/PersonCard/index.tsx
jwinnfeild2517: 👍

ttornese (approved):

PR #841, opened 2022-07-20
WEB-102: Issues with screenreader not reading content

jbonAA:
Change ImageList container from `aside` to `div`
Applies aria-label to sidebar card for better description of complement on tab

* Images without altText will not be focused or read by screen reader
* All valid text and elements can be read on both components using left + right arrows


tdorin (approved):
🌮 🚚

harry-atk (approved):

PR #844, opened 2022-07-25
qa hs jul 25

harry-atk:
- WEB-31: outline-offset no transition for safari
- WEB-34: Hover media on favorite ribbon
- WEB-38: Refinement count color from nobel to eclipse
- WEB-25: Lead Marquee line height 26px
- WEB-27: Current Refinements close icon alignment
- WEB-30: remove focus-within for standard card


harry-atk:
QA passed

PR #845, opened 2022-07-25
FEA-42: Convert RelatedDocumentCard to ts

jbonAA:


src/components/Cards/RelatedDocumentCard/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default } from './RelatedDocumentCard';
+export * from './RelatedDocumentCard';
```
oddwili, line 2: This file isn't doing much. You can probably delete it and just have everything pull from `RelatedDocumentCard.tsx`

oddwili (changes requested):

src/components/Cards/RelatedDocumentCard/index.ts
jbonAA: Sounds good, I'll make that adjustment moving forward!

oddwili (approved):

PR #851, opened 2022-07-26
Hero Ad -> TS; Added Types to /color

RomanTurner:


oddwili (approved):

jbonAA (approved):
Noticed some opportunities to refactor the css on this component to leverage `src/styles/breakpoints.ts` instead of `styled-components-breakpoint`.

ex:
`export const md: BreakpointFn = interp => breakpoint('md')`${interp}`;`
as
`${md(css``)}`

Otherwise 👍

src/components/Ads/HeroAd/HeroAd.tsx
```diff
@@ -14,119 +13,144 @@ import {
...
 const HeroAdContentTheme = {
   default: css`
-    ${breakpoint('xs', 'smmd')`
-      max-width: 20rem;
+  ${sm(css`
+  max-width: 20rem;
       min-width: 18.4rem;
-    `}
-
-    ${breakpoint('xs', 'md')`
```
RomanTurner, line 69: @jbonAA This is what you were referring to? 
I am not super familiar with styled components yet but this doesn't seem like much of a refactor. 
The naming is less intuitive, and the code is often duplicated in spots.

PR #852, opened 2022-07-26
convert herocard to tsx

jwinnfeild2517:
## What does the Pr do

converts the herocard into typescript


src/components/Cards/HeroCard/HeroCard.tsx
```diff
@@ -86,7 +86,7 @@ const HeroCardTitle = styled.div`
   margin-bottom: ${spacing.xxsm};
 `;
 
-const HeroCardDescription = styled.div`
+const HeroCardDescription = styled.div<{ cardType: string }>`
```
ttornese, line 89: seems like `cardType` is one of `'learn' | 'watch'` rather than a generic string, might want to change the typing here to reflect that

src/components/Cards/HeroCard/HeroCard.tsx
```diff
@@ -153,6 +153,26 @@ const HeroCardCtaText = styled.span`
...
+  /** Optional: Key value that maps to a show logo. */
+  iconKey?: string,
+  /** Optional: Image data that is used to render a PersonHeadShot. */
+  personHeadShot: PersonHeadshotPropTypes,
+  /** Optional: Data used to render a sticker. */
+  sticker: PropTypes.InferProps<typeof Sticker.propTypes>,
+  /** Title text above description */
+  title?: string,
+  onClick?: () => void,
+
```
oddwili, line 173: We should remove the comments here. It's clear from the names what the props do and the props that are optional are denoted with a `?`

oddwili (changes requested):

PR #855, opened 2022-07-27
FEA-48 FreeTrialAd -> TS

RomanTurner:
Refactored component to ts and story to ts with use of args/decorators.


.nvmrc
```diff
@@ -1 +1 @@
-14.16.0
```
RomanTurner, line 1: After changing this per Tom's recommendation I have noticed less lag during TS conversion. @oddwili This should help.

ttornese (approved):

PR #866, opened 2022-08-01
FEA-56: Convert ShowMoreLess to ts

jbonAA:


src/components/ShowMoreLess/ShowMoreLess.stories.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const Template: ComponentStory<typeof ShowMoreLess> = (args: ShowMoreLessProps) => (
+  <ShowMoreLess {...args} />
+);
+
+export const Default: ComponentStory<typeof ShowMoreLess> = Template.bind({});
+Default.args = {
+  id: 'something unique',
+  initialCount: 2,
+  items: defaultItems,
```
ttornese, line 28: if these stories are using the `addThemedWrapper` decorator you might want to pass a default siteKey here, otherwise it wont be themed be default

ttornese (approved):
small suggestion, otherwise 👍

PR #889, opened 2022-08-10
convert loading card story to tsx

jwinnfeild2517:
## What does this PR do
update loading card stories for typescript and new story pattern


src/components/Cards/LoadingCard/LoadingCard.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { withThemes, color } from '../../../styles';
+
+export default {
+  title: 'Components/Cards/LoadingCard',
+  component: LoadingCard,
+  argTypes: { mode },
+} as ComponentMeta<typeof LoadingCard>;
+
+const StoryWrapperTheme = {
+  default: css`
```
jbonAA, line 17: Unless you want these specifics styles for a story wrapper, just noting that we have a `DarkModeWrapper` in `mode-dark.tsx` that can be reused

jbonAA (approved):
👍

src/components/Cards/LoadingCard/LoadingCard.stories.tsx
jwinnfeild2517: We need these specific styles so the background matches the loading card mode

PR #894, opened 2022-08-11
eslint set up with prettier and eslint-storybook

RomanTurner:
This configuration will use prettier as our formatter and use eslint as our linter.  To follow -> big linting over the app, and pre-commit hooks with husky to help us keep linting to a standard.


RomanTurner:
The checks will not pass until we do the BIG LINT.

.prettierrc
```diff
@@ -0,0 +1,6 @@
+{
+    "trailingComma": "es5",
+    "tabWidth": 4,
+    "semi": false,
+    "singleQuote": true
+}
```
harry-atk, line 6: should match https://github.com/Americastestkitchen/espresso/blob/WEB-237/.prettierrc#L9-L10 for consistency

PR #897, opened 2022-08-17
core-582: Issue loading TikTok videos on Firefox ATK, CIO, and CCO

tdorin:
# [CORE-582]([SHORTCUT_LINK](https://bostoncommonpress.atlassian.net/browse/CORE-582))

## What Does This Do?
Allows TikTok to re-render on refresh, and to render in all browsers. 

## To view the bug:
visit: https://mise-ui.netlify.app/?path=/story/components-mediaembed--tik-tok and refresh page, or simply select another story and return to the tiktok embed. iframe does not re-render.

Or view https://www.americastestkitchen.com/articles/5741-the-world-s-easiest-peanut-sauce-has-2-ingredients in Firefox

### Author Tested
- [x] Chrome
- [ ] Edge
- [x] Firefox
- [x] Safari

### DIFF
--- a/src/components/MediaEmbed/utilities.ts
+++ b/src/components/MediaEmbed/utilities.ts
@@ -1,13 +1,8 @@
-import { useEffect, useRef, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 export type EmbedProps = { source: string; caption?: string };
 
-/**
- * Attach async script tag. Do nothing when src is undefined.
- * @param src script src to load async.
- * @param onLoad Callback when script has finished loading.
- */
-export const useScript = (src?: string, onLoad?: () => void): void => {
+export const useScript = (src: string, onLoad?: () => void): void => {
   useEffect(() => {
     if (!src) return () => {};
     const script = document.createElement('script');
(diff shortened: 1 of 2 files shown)


harry-atk (commented):
This looks like it adds script tags on every render now. And would do so for the other components that use useScript. Do we just need to wait until we're sure the component mounts before adding the script or is there something else going on?

tdorin:
> This looks like it adds script tags on every render now. And would do so for the other components that use useScript. Do we just need to wait until we're sure the component mounts before adding the script or is there something else going on?

Got it! Per our conversation, I made the update!

oddwili (approved):

src/components/MediaEmbed/utilities.ts
```diff
@@ -1,15 +1,9 @@
...
 
-/**
- * Attach async script tag. Do nothing when src is undefined.
- * @param src script src to load async.
- * @param onLoad Callback when script has finished loading.
- */
-export const useScript = (src?: string, onLoad?: () => void): void => {
+export const useScript = (src: string, onLoad?: () => void): void => {
   useEffect(() => {
-    if (!src) return () => {};
```
harry-atk, line 12: could we leave `if (!src) return () => {};` on useScript.

src/components/MediaEmbed/utilities.ts
```diff
@@ -24,45 +18,34 @@ export const useScript = (src?: string, onLoad?: () => void): void => {
...
+    scr.src = script;
+    document.body.appendChild(scr);
+    return () => {
+      document.body.removeChild(scr);
+    };
+  });
   useEffect(() => {
     fetch(`${baseUrl}${source}`)
-      .then(res => (res.ok ? res.json() : undefined))
+      .then(res => (res.ok ? res.json() : ''))
```
harry-atk, line 44: lets leave this back to undefined to match catch condition

src/components/MediaEmbed/utilities.ts
```diff
@@ -24,45 +18,34 @@ export const useScript = (src?: string, onLoad?: () => void): void => {
...
       .catch(() => undefined)
       .then((payload) => {
-        if (isMounted.current) {
-          setData(payload);
-        }
+        setData(payload.html);
```
harry-atk, line 47: should be `payload?.html` for the conditions that don't return a payload. If you want empty condition to be a string `payload?.html ?? ''`

harry-atk (changes requested):

src/components/MediaEmbed/utilities.ts
tdorin: Good catch!!

harry-atk (approved):

atk-dm:
:rocket: PR was released in [`v1.33.36`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.33.36) :rocket:

PR #916, opened 2022-09-12
task: creating first set of global partials that will be used in table components

ttornese:
Creating first set of global partials for mise 2.0: CircularButton, CircularIcon, and Sticker. Global partials are meant to be used across multiple components in mise 2.0. They will live in `src/components/partials` and should not be rendered on a page directly. The first use case for these will be in the Comparison and Informational table components.

### DIFF
--- a/v2/src/components/partials/CircularButton/CircularButton.tsx
+++ b/v2/src/components/partials/CircularButton/CircularButton.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+
+import CircularIcon from '../CircularIcon/CircularIcon';
+import { IconType } from '../../tokens/icons/svgs/useIconMap';
+import styles from './circularButton.module.scss';
+
+type CircularButtonProps = {
+  iconType: IconType;
+  onClick: () => void;
+};
+
+const CircularButton = ({ iconType, onClick }: CircularButtonProps) => (
+  <button className={styles.button} onClick={onClick}>
+    <CircularIcon type={iconType} />
+  </button>
+);
+
+export default CircularButton;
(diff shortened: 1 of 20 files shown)


v2/src/components/partials/CircularButton/CircularButton.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { IconType } from '../../tokens/icons/svgs/useIconMap';
+import styles from './circularButton.module.scss';
+
+type CircularButtonProps = {
+  iconType: IconType;
+  onClick: () => void;
+};
+
+const CircularButton = ({ iconType, onClick }: CircularButtonProps) => (
+  <button className={styles.wrapper} onClick={onClick}>
```
bradwaropay, line 13: Minor thing, but for things like this consider just naming it something like `button` instead of `wrapper`. Keep things like `wrapper` or `container` for actual surrounding external elements.

v2/src/components/partials/CircularButton/circularButton.module.scss
```diff
@@ -0,0 +1,11 @@
+@import "../../../styles/main.scss";
+
+.wrapper {
+  &:hover {
+    cursor: pointer;
+
+    div {
```
bradwaropay, line 7: We should avoid styling anything by element and always have a class.

v2/src/components/tokens/icons/svgs/CloseIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <svg
+      aria-hidden="true"
+      focusable="false"
+      xmlns="http://www.w3.org/2000/svg"
+      width="12.828"
+      height="12.828"
+      viewBox="0 0 12.828 12.828"
+    >
+      <g transform="translate(1.414 1.414)">
```
bradwaropay, line 15: We should discuss with design about getting these cleaned up and exported properly so there is no inline transforms.

bradwaropay (approved):

PR #930, opened 2022-09-23
feat: ✨ adding temporary brand styles to EditorialTable component

ttornese:
For release editorial tables need to support different styles across our 3 brands. These brand specific styles will go away as we move forward with more brand consolidation starting with the homepage redesign. Since these styles are temporary, a robust theming solution is not needed. I updated the `EditorialTable` component to take a `domainSiteKey` prop. If that prop is present and is not `atk`, a site key module class will be applied to the table that will override button and text colors.

### DIFF
--- a/v2/src/components/blocks/editorialTables/EditorialTable/EditorialTable.stories.tsx
+++ b/v2/src/components/blocks/editorialTables/EditorialTable/EditorialTable.stories.tsx
@@ -238,6 +238,7 @@ ComparisonTablePaginated.args = {
     ],
     type: 'comparison',
   },
+  brandKey: 'atk',
 };
 
 export const ComparisonTableNoPagination = Template.bind({});
@@ -329,6 +330,7 @@ ComparisonTableNoPagination.args = {
     ],
     type: 'comparison',
   },
+  brandKey: 'atk',
 };
 
 export const InformationalTablePaginated = Template.bind({});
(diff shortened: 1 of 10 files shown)


v2/src/components/blocks/editorialTables/EditorialTable/editorialTable.module.scss
```diff
@@ -12,3 +12,72 @@
...
+    :global(.circular-icon) {
+      background-color: $ccoCBPrimary !important;
+
+      &:hover {
+        background-color: $ccoCBPrimaryHover !important;
+      }
+    }
+  }
+
+  th {
```
bradwaropay, line 37: I'd like us to try to avoid styling elements directly and always use some kind of class for scoping purposes within components and identification.

v2/src/components/blocks/editorialTables/EditorialTable/EditorialTable.stories.tsx
```diff
@@ -511,4 +514,5 @@ InformationalTableNoPagination.args = {
     ],
     type: 'informational',
   },
+  domainSiteKey: 'atk',
```
bradwaropay, line 517: Nit picky but maybe change this to something like brandKey or similar as we are dropping multiple domains.

bradwaropay (changes requested):

bradwaropay (approved):

PR #934, opened 2022-09-29
feat: ✨ creating EditorialText component to be used to render WSYIWYG content from barista

ttornese:
mise v2.0 update

* created `EditorialText` component that should be used to render WYSIWYG content from barista
* Component currently supports styling for links, italicized, and bolded text (ordered and unordered list support should be added later)
* Supports rendering with primary font (currently proximaNova, but may change) and accent font (currently merriweather)
* Font sizes and line heights should be set by the parent component that renders EditorialText as those values vary for WYSIWYG content across pages

### DIFF
--- a/v2/.storybook/preview-head.html
+++ b/v2/.storybook/preview-head.html
@@ -1,0 +1,1 @@
+<link href="https://fonts.googleapis.com/css?family=Merriweather:400,400i,700&display=swap" rel="stylesheet">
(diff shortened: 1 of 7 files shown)


tdorin (approved):

atk-dm:
:rocket: PR was released in [`v1.33.40`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.33.40) :rocket:

PR #935, opened 2022-09-30
feat: :sparkles: Recipe Header Component

tdorin:
Developing the recipe Component Header. this includes creating the Global Partial for the BylineList and stories, as well as the scoped partials to create the Header


v2/src/components/blocks/recipeComponents/partials/RecipeHeader/RecipeHeader.tsx
```diff
@@ -0,0 +1,40 @@
...
+const RecipeHeader = ({
+    authors,
+    attribution,
+    header,
+    onClick
+  }: RecipePropTypes) => {
+    const {yields, time, name, headnote} = header
+    return (
+    <div className={styles.recipeHeaderWrapper}>
+        <h2 className={styles.recipeHeadline}>{name}</h2>
```
ttornese, line 23: since you're importing these styles for a module, the class names will all be locally scoped to this component. `recipeHeader` (or whatever the module name is in general) will be prepended to every class you import from that module. you can shorten the class names here by dropping `recipeHeader` from them and just using names like `styles.wrapper` and `styles.headline`. when these components are in the DOM the class names will appear like so: `class="recipeHeader_wrapper-some-random-hash"`

v2/src/components/blocks/recipeComponents/partials/RecipeHeader/recipeHeader.module.scss
```diff
@@ -0,0 +1,26 @@
+@import "../../../../../styles/main.scss";
+
+.recipeHeaderWrapper {
+    .recipeHeadline {
```
ttornese, line 4: Since module classes are locally scoped, you don't need to nest other module classes within the `wrapper` module class. Only the classes selected with the `:global` selector need to be nested.

v2/src/components/blocks/recipeComponents/types.ts
```diff
@@ -0,0 +1,17 @@
...
+  export type Author = {
+    id: number;
+    firstName: string;
+    lastName: string;
+    image?: { altText?: string, cloudinaryUrl: string };
+    inactive?: boolean;
+  };
+  
+  export type Header = {
+    recipeId: string;
```
ttornese, line 12: It looks like this id isn't used anywhere aside from this type definition. Can it be dropped?

v2/src/components/partials/Byline/bylineList.module.scss
```diff
@@ -0,0 +1,48 @@
+@import "../../../styles/main.scss";
+
+.bylineWrapper {
```
ttornese, line 3: Same note as I made above for the recipe header module styles, `byline` can be dropped for all of these class name selectors

v2/src/components/partials/Byline/bylineList.module.scss
```diff
@@ -0,0 +1,48 @@
...
+    }
+    .personHeadShot { 
+        flex-shrink: 0;
+        overflow: hidden;
+        border-radius: 50%;
+        margin-right: 12px;
+        height: 40px;
+        width: 40px;
+    }
+    img {
```
ttornese, line 29: Can we create a module class name selector for this image instead of using the generic img el selector?

v2/src/components/blocks/recipeComponents/partials/RecipeHeader/RecipeHeader.tsx
tdorin: Whoa, Good call!

PR #936, opened 2022-10-04
feat: :sparkles: Recipe Component Ingredient List

tdorin:
Building out the Ingredient List within the Recipe Component.

### DIFF
--- a/v2/src/components/blocks/recipeComponents/partials/RecipeIngredientsList/RecipeIngredientsList.stories.tsx
+++ b/v2/src/components/blocks/recipeComponents/partials/RecipeIngredientsList/RecipeIngredientsList.stories.tsx
@@ -1,0 +1,142 @@
+import React from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+
+import RecipeIngredientsList from './RecipeIngredientsList';
+
+export default {
+  title: 'Components/Blocks/RecipeComponent/RecipeIngredientsList',
+  component: RecipeIngredientsList,
+} as ComponentMeta<typeof RecipeIngredientsList>;
+
+const Template: ComponentStory<typeof RecipeIngredientsList> = args => <RecipeIngredientsList {...args} />;
+
+const ingredientListObject = [
+  {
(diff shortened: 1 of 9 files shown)


v2/src/components/blocks/recipeComponents/partials/RecipeIngredientsList/RecipeIngredientsList.stories.tsx
```diff
@@ -0,0 +1,160 @@
...
+} as ComponentMeta<typeof RecipeIngredientsList>;
+
+const Template: ComponentStory<typeof RecipeIngredientsList> = args => <RecipeIngredientsList {...args} />;
+
+const ingredientListObject = [
+  {
+    name: "This is the First Ingredient Group Name",
+    items: [
+      {
+        pre: "",
```
ttornese, line 18: The pre-text field is going to be removed from Barista since editorial no longer wants to use it, so we can drop this field from the story, component, and types.

v2/src/components/blocks/recipeComponents/partials/RecipeTable/RecipeTable.tsx
```diff
@@ -0,0 +1,30 @@
...
+const RecipeTable = ({pre, post, quantity, measurement, name}: IngredientItems) => {
+    return(
+      <div>
+        <label className={styles.checkbox}>
+          <input 
+          className={styles.checkboxInput}
+          id={`ingredient-${name}`}
+          name={`${name}`}
+          type="checkbox"></input>
+          <div className={styles.checkboxIndicator}></div>
```
ttornese, line 16: this div and the input above could be self closing since they dont have child content `<input ... />`

v2/src/components/blocks/recipeComponents/partials/RecipeIngredientsList/RecipeIngredientsList.stories.tsx
tdorin: For real?!!

ttornese (approved):

PR #938, opened 2022-10-05
feat: :sparkles: web-532: recipe component Instructions list

tdorin:
built out the component for the list of instructions within the recipe component


v2/src/components/blocks/recipeComponents/partials/RecipeInstructions/RecipeInstructions.tsx
```diff
@@ -0,0 +1,52 @@
...
+  let count = 0
+  return (
+    <div className={styles.wrapper}>
+      <div className={styles.group}>
+        <h2 className={styles.headline}>Instructions</h2>
+        <EditorialText content={header} fontStyle={'accent'}></EditorialText>
+      </div>
+      {
+        list.map((item) => {
+          count = count + 1
```
ttornese, line 25: `count` here seems to be the same value as `index + 1`. you could use the array index to keep track of count instead: `list.map((item, index))`

PR #991, opened 2023-06-01
web-2539 Art19 Token 

tdorin:
the Stories for Audio Player, Podcast Episode & Show More Results, all exposed the Art19 API credentials in the header fetch function. Since the tokens were compromised, we have new ones assigned, but they cannot be used in any of the Mise Stories. As this will be sunset anyway, we are simply deleting the stories that exposed the token in Github.


ttornese (approved):

atk-dm:
:rocket: PR was released in [`v1.33.50`](https://github.com/Americastestkitchen/mise-ui/releases/tag/v1.33.50) :rocket:

